PR #5, opened 2019-01-22
Documentation template

lucperkins:
@mfdii I think this is ready for review. It's nothing too fancy at the moment but definitely something we can iterate on. I haven't yet transferred all of the existing material over from the wiki but I think there's enough material there to get a sense of the overall aesthetic and such.

You can see the documentation here:

https://deploy-preview-5--falcosecurity.netlify.com/docs/


mfdii:
The only suggestion I have is to have the header match what's on the home page. The header currently comes back when you begin scrolling, but if you drop directly into `/docs`, and don't scroll you're a bit "stuck".

lucperkins:
@mfdii Good call on that. I can add a "sticky" top navbar for the docs, which should clear up that issue.

mfdii:
One other thing I thought of as I was reviewing this: We need to be able to have a table of contents for different sections. For example, I'd like to have a quickstart guide and that should be one of the main menu items along the top menu bar. When a user clicks on the quickstart, it should take them to the main page of the quickstart, and on that page along the left we should have a TOC that lets users jump to different sections of the doc. 

For other pages I'd like to have a TOC that lets us organize the docs and allow users to jump between pages easier. This is a problem we currently have with GitHub wiki. We can't change the "Pages" section that is present on every wiki page, and when a user clicks on a page they "get lost" in the documentation at that point. 

[Kubernetes docs](https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/) is a good example of what I am thinking.  We don't necessarily need to have the menu of the left collapse/expand like k8s docs does.

mfdii:
@lucperkins is this ready to merge? The one thing I didn't see on the netlify preview was the Falco site header is still not showing until you scroll.

mfdii:
Ignore my comment. I double checked and the header is there

lucperkins:
@mfdii Okay, I think this is in reviewable shape. The wiki content has been moved. The aesthetics could use some work but I think it's in solid shape to iterate on later.

PR #11, opened 2019-04-05
updates from wiki before removing wiki

mfdii:
The docs diverged from the wiki. Updating the docs on the site to match the wiki so we can shut down the wiki.

### DIFF
--- a/content/docs/event-sources/dropped-events.md
+++ b/content/docs/event-sources/dropped-events.md
@@ -34,4 +34,4 @@ To reduce the likelihood of a flood of log messages/alerts, actions are governed
 
 ## Configuration
 
-The actions to take on a dropped syscall event and the throttling parameters for the token bucket are configurable in `falco.yaml` and described [[here|Falco-Configuration#syscall_event_drops]].
+The actions to take on a dropped syscall event and the throttling parameters for the token bucket are configurable in `falco.yaml` and described [here](/docs/configuration/#syscall_event_drops).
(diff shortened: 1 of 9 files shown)


content/docs/rules/_index.md
```diff
@@ -76,7 +76,7 @@ As noted above, macros provide a way to define common sub-portions of rules in a
 
 With this macro defined, we can then rewrite the above rule's condition as `in_container and proc.name = bash`.
 
-For many more examples of rules and macros, please take a look `rules/falco_rules.yaml`.
```
lucperkins: You can drop the `.html` in link URLs. Not necessary, just FYI.

mfdii: Thanks for the catch. I did that with others, but missed that one.

content/docs/event-sources/dropped-events.md
```diff
@@ -0,0 +1,37 @@
...
+Wed Mar 27 15:28:24 2019: Exiting.
+```
+
+## Actions Rate Throttling
+
+To reduce the likelihood of a flood of log messages/alerts, actions are governed by a token bucket. The default parameters of the token bucket are a rate of one alert per 30 seconds with an allowed burst of up to 10 messages.
+
+## Configuration
+
+The actions to take on a dropped syscall event and the throttling parameters for the token bucket are configurable in `falco.yaml` and described [[here|Falco-Configuration#syscall_event_drops]].
```
lucperkins: One more thing: this link should be `[here](/docs/configuration/#syscall_event_drops)`. GitHub wiki syntax is unfortunately not supported in standard Markdown.

mfdii: Nice catch. Just missed updating it.

PR #14, opened 2019-04-26
Add tester docs

mstemm:


mstemm:
@mfdii can you look at the falco-builder/falco-tester sections to make sure they explain the process in sufficient detail? Then we'lll merge.

Issue #30, opened 2019-06-19
Right sections sub menu indentation


fntlnz:
While working on the build documentation, we noticed that the right menu doesn't respect indentation for sub-sections, this creates confusion on the user while trying to figure out what's the hierarchy of the sections.

Here's how the menu looks like now, highlighted in yellow:

<img width="1637" alt="image" src="https://user-images.githubusercontent.com/3083633/59749870-1d806880-927e-11e9-8e64-7701be4fd4ee.png">

Here's how it should be, respecting the sub-sections:
<img width="311" alt="image" src="https://user-images.githubusercontent.com/3083633/59750135-97b0ed00-927e-11e9-8998-3add7f53dce5.png">

PR #32, opened 2019-06-20
Add indented TOC

lucperkins:
Addresses issue #30

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,7 @@ dependencies:
 serve: dependencies
 	hugo server \
 		--buildDrafts \
-		--buildFuture \
-		--disableFastRender
+		--buildFuture
 
 production-build: dependencies
 	hugo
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -4,8 +4,7 @@ dependencies:
 serve: dependencies
 	hugo server \
 		--buildDrafts \
-		--buildFuture \
-		--disableFastRender
+		--buildFuture
```
leodido: The hugo docs state: `buildFuture` doesn't set which content is scheduled for the future. That's the task of the `publishdate`. When that front matter variable is set to a date in the future, Hugo won't render the content. And when `publishdate` is in the past (relative to the current date), the content gets rendered.

I'm asking just to ensure it is the behaviour we want :)

leodido:
We are going to remove the example file with another PR right now.

PR #33, opened 2019-06-21
fix: remove leftover example

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

Co-Authored-By: Leonardo Di Donato <leodidonato@gmail.com>


lucperkins:
@fntlnz @leodido FYI this was unnecessary, as that document is labeled as a draft and thus doesn't show up on the production site.

PR #91, opened 2019-10-07
Use backtick-based highlighting

lucperkins:
There's nothing wrong with using Hugo's `{{< highlight >}}` shortcode for syntax highlighting, and there are even cases where it's necessary, but when you do so you lose (a) text editor compatibility, and (b) highlighting on GitHub.


fntlnz:
Thanks @lucperkins this produces the same output so if we get all the other benefits even better!

PR #92, opened 2019-10-15
wip: probes list

Issif:
Signed-off-by: Issif <issif+github@gadz.org>

**What type of PR is this?**

/kind content

/area documentation

This PR provides a list of all available probes in S3. The list is formatted into several tables, with a search box to filter.

N/A

This PR is not complete right now, the sources for lists are still on my side : http://thomas.labarussias.fr/falco-probes/data/. We need to figure out how we can generate these json with Netlify or else.


fntlnz:
@Issif any blocker on this?

lucperkins:
I think it would be much easier to statically generate the table using Hugo rather than fetching JSON

Issif:
@fntlnz for now, JSON are still on my personnal account, we need to find how and where to generate them

@lucperkins I did it in purpose, because I would these JSON be used by other tools like falcoctl (`falco search probe` which search in JSONs if the adapted probe is available). Moreover, probes are added from time to time, no need to rebuild website each time.

PR #108, opened 2019-12-23
Reword program output doc and add an example

marier-nico:
**What type of PR is this?**

/kind content

/area documentation

This PR rewords the introduction for the "program output" section of the documentation. It is slightly less concise now, but readability has been improved. Also, an example was added for using `xargs` with program outputs. This is useful when the program to use does not read from `stdin`.

There are no issues associated with this PR!


kris-nova:
/approve

Thanks for your contribution!

PR #130, opened 2020-02-07
Make docs content narrower

lucperkins:
This PR restricts the length of lines in written text to 90 characters, which should provide a readability improvement.


lucperkins:
@leodido I'll see if I can make it a bit wider for very large displays, but in general lines should never be preferably 80 characters or less and pretty much never over 100, even if it looks a bit silly on a display like that. UX research into readability suggests these limits.

PR #151, opened 2020-02-25
Move _redirects file into static directory

lucperkins:
The `_redirects` file for Netlify is not currently being applied because it is not part of the website "payload." This PR moves it into the `static` directory so that it ccan be properly recognized and used by the publishing platform.


lucperkins:
@leodido I think it needs to be LGTM to approve 😄

leodido:
I clicked comment rather than approve LOL

PR #162, opened 2020-03-08
new(blog): add content for minikube support for Falco

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

**What type of PR is this?**

/kind content

/area blog

Fixes #

### DIFF
--- a/content/en/blog/minikube-falco-kernel-module.md
+++ b/content/en/blog/minikube-falco-kernel-module.md
@@ -1,0 +1,142 @@
+---
+title: Minikube 1.8.0 packages the Falco Kernel Module
+date: 2020-03-08
+author: Lorenzo Fontana
+---
+
+Minikube is a tool that implements a local Kubernetes cluster on macOS, Linux and Windows via a simple command line, it is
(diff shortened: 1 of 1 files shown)


content/en/blog/minikube-falco-kernel-module.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+Read the logs to make sure Falco is running!
+
+
+```bash
+kubectl logs -l app=falco -f
+```
+
+It will show errors trying to find the Kernel headers, those are needed to build the module but we can just skip those errors because the module is already there!
```
kris-nova: Can we update the Falco messaging to reflect this? I think we should be able to detect if we are running in Minikube and load accordingly.

content/en/blog/minikube-falco-kernel-module.md
```diff
@@ -0,0 +1,142 @@
...
+
+```
+* Unloading falco-probe, if present
+* Running dkms install for falco
+Error! echo
```
kris-nova: See my note above - if we aren't erroring - we shouldn't error - right?

PR #206, opened 2020-05-20
update: direct link to the new slack channel

leogr:
**What type of PR is this?**

/kind content

/area blog

/area documentation

Quoting @leodido who explained the reason very well: 

> After a [very long discussion](https://github.com/falcosecurity/falco/issues/983) with CNCF, we finally decided to move the Slack community over the Kubernetes slack (as per Chris A. 's suggestion [here](https://github.com/falcosecurity/falco/issues/983#issuecomment-625901549)).
> 
> We are going to [officially announce the cutoff](https://github.com/falcosecurity/falco/issues/983#issuecomment-630980624) during today community call! 
> 
> And we already started putting documentation about the new channels (and the new slack) in place for the community (see [here](https://github.com/falcosecurity/community/pull/95/files)).
> 

Fixes #

The #falco channel on Kubernetes Slack was added [here](https://github.com/kubernetes/community/pull/4038). So we can update it right now.

/cc @kris-nova 
/cc @leodido 
/cc @fntlnz

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -90,7 +90,7 @@ languages:
         subtitle: Sign up for the development mailing list
         buttonText: Submit
         slackText: Falco Slack Channels
-        slackLink: https://slack.k8s.io/
+        slackLink: https://kubernetes.slack.com/archives/falco
         mailingListText: Developer Mailing List
         mailingListLink: https://lists.cncf.io/g/cncf-falco-dev
         communityText: Github Community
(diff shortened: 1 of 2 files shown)


fntlnz:
I don't think that what we are doing here is right. The original link points to the signup page for the k8s slack channel while the updated links point to the specific channel. If you are not registered, you have no way to get into without the first link.

leogr:
> I don't think that what we are doing here is right. The original link points to the signup page for the k8s slack channel while the updated links point to the specific channel. If you are not registered, you have no way to get into without the first link.

If you're not registered, slack will redirect you to the sign in page, then to the channel. So what's wrong?

leogr:
oh, maybe I have noticed the problem
/hold

leogr:
Users need an invite to join the k8s slack, so we have to keep the k8s slackin site until we find a solution.

Closed for now.

PR #382, opened 2021-01-26
docs(content/en/docs/rules): add warning for `Via Custom Rule Definion` section

leogr:
See also: https://github.com/falcosecurity/falco/issues/1537

Signed-off-by: Leonardo Grasso <me@leonardograsso.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

See https://github.com/falcosecurity/falco/issues/1537

Fixes #

/cc @Rajakavitha1

### DIFF
--- a/content/en/docs/rules/_index.md
+++ b/content/en/docs/rules/_index.md
@@ -215,9 +215,11 @@ Should `proc.name=nginx` be interpreted as relative to the `and proc.name=apache
 In cases like this, be sure to scope the logical operators of the original condition with parentheses when possible, or avoid appending conditions when not possible.
 
 ## Disable Default Rules
+
 Even though Falco provides a quite powerful default ruleset, you sometimes need to disable some of these default rules since they do not work properly in your environment. Luckily Falco offers you multiple possibilities to do so.
 
 ### Via existing Macros
+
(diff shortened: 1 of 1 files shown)


content/en/docs/rules/_index.md
```diff
@@ -251,11 +251,20 @@ Last but not least, it's also possible to just disable a by default enabled rule
   enabled: false
 ```
 
-This is especially useful for rules which do not provide a `consider_*` macro in it's default condition. Again, ensure the falco configuration files are loaded in the right order.
+This is especially useful for rules which do not provide a `consider_*` macro in it's default condition. Again, ensure the Falco configuration files are loaded in the right order.
```
Rajakavitha1, line 254: How about?
Last but not the least, you can just disable a rule that is enabled by default using a combination of the `append: true` and `enabled: false` rule properties.
This is especially useful for rules which do not provide a `consider_*` macro in the default condition.

Ensure that the custom configuration file loads after the default configuration file. You can configure the right order using multiple `-r` parameters, directly inside the falco configuration file `falco.yaml` through `rules_file`. If you are using the official Helm chart, then configure the order with the `falco.rulesFile` value.

For example to disable the `User mgmt binaries` default rule in `/etc/falco/falco_rules.yaml` define a custom rule in `/etc/falco/rules.d/custom-rules.yaml`:

```yaml
- rule: User mgmt binaries
  append: true
  enabled: false
```


 {{% pageinfo color="warning" %}}
 There appears to be a bug with this feature that we are looking into. If `enabled: false` doesn't work, you can use the following workaround as an alternative:
 ```yaml
 - rule: User mgmt binaries
   condition: and (never_true)
   append: true
 ```
 {{% /pageinfo %}}

leogr: Thank you! much better

leodido (approved):

PR #430, opened 2021-05-13
Remove repeated text

ramonssarmento:

Remove repeated text on  "Load latest falco kernel module"


leogr (approved):
Good catch!
Thank you!

PR #489, opened 2021-08-25
content(rules): update events, conditions and fields documentation 

lucklypse:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

This PR updates the rules part of the documentation with the following:
- Some documentation about the condition language was missing from the Falco website. A brand new stand alone page is added for it in `conditions.md`
- Formatting for the event field pages is changed to be easier to read on the falco website (see cosmetic change in the screenshot below)
- A page, `supported-events.md` is introduced to list the current supported events
- The Rules page is updated to reflect recent changes and to link the pages above

image

Fixes #

Note that the `--markdown` options that are used to generate the events and fields list is not available yet in stock Falco. Simple modifications to libs and falco are being submitted to reflect this format.


content/en/docs/rules/_index.md
```diff
@@ -48,38 +48,51 @@ Key | Required | Description | Default
...
 
 ```
 container.id != host and proc.name = bash
 ```
 
-The first clause checks that the event happened in a container (Sysdig events have a `container` field that is equal to `"host"` if the event happened on a regular host). The second clause checks that the process name is `bash`. Note that this condition does not even include a clause with a system call! It only checks event metadata. Because of that, if a bash shell does start up in a container, Falco outputs events for every syscall that is performed by that shell.
+The first clause checks that the event happened in a container (where `container.id` is equal to `"host"` if the event happened on a regular host). The second clause checks that the process name is `bash`. Since this condition does not include a clause with a system call it will only check event metadata. Because of that, if a bash shell does start up in a container, Falco outputs events for every syscall that is performed by that shell.
 
-A complete rule using the above condition might be:
+If we only want to be alerted for each successful spawn of a shell in a container we can add the appropriate event type and direction to our condition:
```
radhikapc, line 61: If you want to be alerted only for each successful spawn of a shell in a container, add the appropriate event type and direction to our condition:

content/en/docs/rules/_index.md
```diff
@@ -48,38 +48,51 @@ Key | Required | Description | Default
...
+
+```
+evt.type = execve and evt.dir=< and container.id != host and proc.name = bash
+```
+
+And so, a complete rule using the above condition might be:
```
radhikapc, line 67: Therefore, a complete rule using the above condition might be:

content/en/docs/rules/_index.md
```diff
@@ -48,38 +48,51 @@ Key | Required | Description | Default
...
 ```yaml
 - rule: shell_in_container
   desc: notice shell activity within a container
-  condition: container.id != host and proc.name = bash
+  condition: evt.type = execve and evt.dir=< and container.id != host and proc.name = bash
   output: shell in a container (user=%user.name container_id=%container.id container_name=%container.name shell=%proc.name parent=%proc.pname cmdline=%proc.cmdline)
   priority: WARNING
 ```
 
+Conditions allow us to check for many aspects of each supported event. To learn more, check out our more in-depth discussion about the [condition language](/docs/rules/conditions).
```
radhikapc, line 77: Conditions allow you to check for many aspects of each supported event. To learn more, see the [condition language](/docs/rules/conditions).

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Condition Syntax
+weight: 3
+---
+
+A condition is a boolean expression related to a single event that has been detected by Falco. You can use fields related to every supported event, but we'll focus on syscalls as they're currently the most common. The language supports boolean operators and parentheses as you'd expect. For example a condition like:
```
radhikapc, line 6: A condition is a boolean expression related to a single event that has been detected by Falco. You can use fields related to every supported event, but this document focuses on syscalls as they're currently the most common. The language supports boolean operators and parentheses as you'd expect. For example a condition like:

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+evt.type = execve and evt.dir = < and (proc.name = cat or proc.name = grep)
+```
+
+Will trigger for each execution of `cat` or `grep`. We'll now take a closer look at what is the meaning of those fields such as `evt.dir`, how to discover the types of events there are available and which fields go with which event type.
```
radhikapc, line 12: This will trigger for each execution of `cat` or `grep`.You can take a closer look at what is the meaning of those fields such as `evt.dir`, how to discover the types of events available, and which fields go with which event type.

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Syscall event types, direction and args
+
+Every syscall event will present the [`evt` field class](/docs/rules/supported-fields/#field-class-evt). Each condition that we write for those events will normally starts with a `evt.type` expression or macro; this makes a lot of sense since security rules normally consider one syscall type at a time. For instance, we may want to consider `open` or `openat` to catch suspicious activity when opening files, `execve` to inspect spawned processes and so forth. We don't have to guess the syscall name, as we can look at the complete [list of supported system calls events](/docs/rules/supported-events) or to understand which ones we can use.
```
radhikapc, line 16: Every syscall event will present the [`evt` field class](/docs/rules/supported-fields/#field-class-evt). Each condition that you write for those events will normally start with a `evt.type` expression or macro; this makes a lot of sense since security rules normally consider one syscall type at a time. For instance, you may want to consider `open` or `openat` to catch suspicious activity when opening files, `execve` to inspect spawned processes, and so forth. You don't have to guess the syscall name, as you can see the complete [list of supported system calls events](/docs/rules/supported-events) and understand which ones you can use.

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+Every syscall has an entry event and an exit event which is shown in the `evt.dir` field, also known as the "direction" of the system call. A value of `>` indicates entry, which fires when the syscall is invoked, while `<` marks exit, meaning that the call has returned. In fact, by looking at the supported system call list we can see that our events have both entries. For example:
```
radhikapc, line 18: Every syscall has an entry event and an exit event which is shown in the `evt.dir` field, also known as the "direction" of the system call. A value of `>` indicates entry, which fires when the syscall is invoked, while `<` marks exit, meaning that the call has returned. In fact, by looking at the supported system call list you can see that our events have both entries. For example:

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+> setuid(UID uid)
+< setuid(ERRNO res)
+```
+
+Most of the times the engine will tell us a lot more information about exit events as we want to understand what happened after the event execution is complete as we can see with events concerning opening files.
```
radhikapc, line 25: Most of the time the engine informs you about exit events as you want to understand what happened after the event execution is complete. You can see by using the events associated with opening files.

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+> open()
+< open(FD fd, FSPATH name, FLAGS32 flags, UINT32 mode, UINT32 dev)
+> openat()
+< openat(FD fd, FD dirfd, FSRELPATH name, FLAGS32 flags, UINT32 mode, UINT32 dev)
+```
+
+As we can see, each event has a list of arguments associated with it that we can access by using `evt.arg.<argname>`. For example, if we want to identify when a process opens a file to overwrite it we need to check if the list of flags contains `O_TRUNC` we will use the `evt.arg.flags` field of the `open` and `openat` exit events shown above and our rule will then look like:
```
radhikapc, line 34: As you can see, each event has a list of arguments associated with it that you can access by using `evt.arg.<argname>`. For example, to identify when a process opens a file to overwrite it, you need to check if the list of flags contains `O_TRUNC`.  You can use the `evt.arg.flags` field of the `open` and `openat` exit events shown above and the rule will then look like this:

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+evt.type in (open, openat) and evt.dir = < and evt.arg.flags contains O_TRUNC
+```
+
+Note that the arguments do not necessarily match the raw parameters that are used in the Linux kernel, but are parsed by Falco to make it easier to write rules. By using the [`evt` fields](/docs/rules/supported-fields/#field-class-evt) we can inspect many more aspects that are common across events.
+
+# Syscall event context and metadata
+
+While the `evt` fields allow us to write pretty expressive conditions, arguments and common fields are usually not enough to write full security rules. Many times we want to add conditions based on the process context the event happens in, or whether or not something is happening inside a container or even correlate each event with the relevant Kubernetes metadata for our cluster, pods, and more. For this reason Falco enriches many events with [other field classes](/docs/rules/supported-fields). Not all classes are available for all events and the list can grow. The documentation for each clarifies when those are expected to be available, but some are so common that we often rely on them.
```
radhikapc, line 44: While the `evt` fields allow us to write pretty expressive conditions, arguments and common fields are usually not enough to write full security rules. Many times you want to add conditions based on the process context the event happens in, or whether or not something is happening inside a container or even correlate each event with the relevant Kubernetes metadata for the cluster, pods, and more. For this reason, Falco enriches many events with [other field classes](/docs/rules/supported-fields). Not all the classes are available for all the events and the list can grow. The documentation for each clarifies when those are expected to be available, but some are so common that you often rely on them.

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+The [`proc` field class](/docs/rules/supported-fields/#field-class-process) gives us context about the process and thread that is generating a specific syscall. This is usually very important. The most basic piece of information we can get out of it is `proc.name` and `proc.pid`, but we can even traverse the process hierarchy by using `proc.aname[<n>]` and `proc.apid[<n>]`. Likewise, we may be interested in which user performs a specific action via the `user` field class.
```
radhikapc, line 46: The [`proc` field class](/docs/rules/supported-fields/#field-class-process) gives you the context about the process and thread that is generating a specific syscall. This information is usually very important. The most basic piece of information you can get out of it is `proc.name` and `proc.pid`, but you can even traverse the process hierarchy by using `proc.aname[<n>]` and `proc.apid[<n>]`. Likewise, you may be interested in which user performs a specific action via the `user` field class.

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+By going through the documentation we can see for example how to catch executions of `bash` within containers:
```
radhikapc, line 48: The documentation gives you an example of how to catch executions of `bash` within containers:

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+if evt.type = execve and evt.dir = < and container.id != host and proc.name = bash
+```
+
+Note that we didn't even have to look at the `execve` args. That is because once `execve` has returned the process context recorded by Falco is updated, meaning that the `proc.` fields will already refer to all information, including the commandline, executable, arguments, related to the new process that was just spawned.
```
radhikapc, line 54: Note that you don't even have to look at the `execve` args. That is because once `execve` has returned the process context recorded by Falco is updated, meaning that the `proc.` fields will already refer to all information, including the command line, executable, arguments, related to the new process that was just spawned.

content/en/docs/rules/conditions.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Operators
+
+We can use the following operators in Falco conditions: `=`, `!=`, `<=`, `<`, `>=`, `>`, `contains`, `icontains`, `startswith`, `endswith`, `glob`, `in`, `intersects`, `pmatch`, `exists`.
```
radhikapc, line 58: You can use the following operators in Falco conditions: `=`, `!=`, `<=`, `<`, `>=`, `>`, `contains`, `icontains`, `startswith`, `endswith`, `glob`, `in`, `intersects`, `pmatch`, `exists`.

radhikapc (changes requested):
Minor changes

lucklypse:
Thanks a lot for the review @radhikapc . I addressed your comments, and also added a few lines at the end of the conditions section for operators.

radhikapc (approved):
LGTM

PR #493, opened 2021-09-02
Add plugins docs

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

Add documentation for how plugins work and a developer's guide for authoring plugins.

Fixes #


leogr:
/hold

Feel free to cancel the "hold" once the plugin feature has been released!

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+With each new release, make sure the contact information provided by the plugin is up-to-date.
+
+## Go Plugin SDK Walkthrough
+The [Go SDK](https://github.com/falcosecurity/plugin-sdk-go) provides prebuilt constructs and definitions that help developing plugins by abstracting all the complexities related to the bridging between the C and the Go runtimes. The Go SDK takes care of satisfying all the plugin framework requirements without having to deal with the low-level details, by also optimizing the most critical code paths.
+
+This section documents the Go SDK at a high-level, please refer to the [official Go SDK documentation](https://github.com/falcosecurity/plugin-sdk-go) for deeper details.
+
+### Architecture of the Go SDK
+
+Since Falcosecurity plugins run in a C runtime, the Go SDK has been designed to abstract most of the complexity related to writing C-compliat code acceptable by the plugin framework, so that developers can focus on writing Go code only.
```
mstemm, line 99: I think you meant C-compliant?

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+
+With every release, you should check for an updated Plugin API Version and if needed, update the plugin to conform to the new API. Remember that a plugin and framework are considered be compatible if their major versions are the same.
+
+With each new release, make sure the contact information provided by the plugin is up-to-date.
+
+## Go Plugin SDK Walkthrough
+The [Go SDK](https://github.com/falcosecurity/plugin-sdk-go) provides prebuilt constructs and definitions that help developing plugins by abstracting all the complexities related to the bridging between the C and the Go runtimes. The Go SDK takes care of satisfying all the plugin framework requirements without having to deal with the low-level details, by also optimizing the most critical code paths.
+
+This section documents the Go SDK at a high-level, please refer to the [official Go SDK documentation](https://github.com/falcosecurity/plugin-sdk-go) for deeper details.
+
```
mstemm, line 96: I know I've mentioned this in other places like the godoc docs, and you kind of mention this in the next paragraph, but I'd suggest mentioning here that a sdk user can either choose from a lower level set of abstractions or just take the streamlined set of packages, where the "easy" route just involves importing a few packages. To me this will help a new reader avoid being overwhelmed by the packages and know they can just choose a streamlined interface, if they want.

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+#### Defining Extractor Plugins
+In the Go SDK, an extractor Plugin is a type implementing the following interface:
+```go
+// sdk/plugins/extractor
+type Plugin interface {
+	...
+	Info() *sdk.Info
+	Init(config string) error
+	// Destroy()
```
mstemm, line 132: Are the commented out methods intended to be optional? If so, we should mention that in the code block too. And since the notion of "optional" methods on an interface doesn't really exist, maybe we should consider all of these mandatory and deal with the small additional cost of having the plugin author write do-nothing modules rather than having these optional methods not really be part of the interface at all.

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+
+### Getting Started
+The SDK is built on top a set of minimal composable interfaces describing the behavior of plugins and plugin instances. As such, developing plugins is as easy as defining a struct type representing the plugin itself, ensuring that the mandatory interface methods are defined on it, and then registering it to the SDK.
+
+To use the Go SDK, all you need to import are the `sdk` and `sdk/plugins` packages. The first contains all the core types and definitions used across the rest of the SDK packages, whereas the latter contains built-in constructs to develop plugins. The subpackages `sdk/plugins/source` and `sdk/plugins/extractor` contain specialized definitions for source and extractor plugins respectively.
+
+The `dummy` Go plugin, documented in the [next sections](#example-go-plugin-dummy), is a simple example that helps understanding how to start writing Go plugins with this SDK.
+
+#### Defining Extractor Plugins
```
mstemm, line 124: When initially reading this section, it made sense--your plugin is either a source or extractor plugin, which each having its own package and interface.

But in the dummy plugin walkthrough, the interfaces (and packages) are more about functionality--a source plugin that extracts fields uses the sdk/plugins/extractor package and interface too.

Can we mention this distinction here as well as in the dummy plugin walkthrough?

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+
+`String()` returns a string representation of an event created by the plugin. The string representation should be on a single line and contain important information about the event. 
+
+The `source.Instance` interface represents plugin instances for an opened event stream, and has one mandatory method and few optional ones.
+
+`NextBatch` creates a new batch of events to be pushed in the event stream. The SDK provides a pre-allocated batch to write events into, in order to manage the used memory optimally.
+
+`Close()` is an optional method that can be used to release the resources used by a plugin instance before its event stream gets closed.
+
+`Progress()` is an optional method returning a float and string representation of the current progress of the event stream. This applies to all the event stream for which the total number of events can be determined.
```
mstemm, line 183: I would mention the units here. Is it 0-100% and the values are converted back to 0-10000 as the underlying API uses, or should the values be 0-10000 here (in which case they should probably be an int, not float).

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+	log.Printf("[%s] Close\n", PluginName)
+}
+```
+
+#### Returning Events
+
+New events are generated in batch by the `NextBatch` function. The function is mandatory for source plugins and must be defined as a method of the plugin instance struct type. The `pState` argument is the plugin struct type initialized in `Init()`, passed in by the framework for ease of access. The plugin state is passed as an instance of the `sdk.PluginState` interface, so a manual cast is required to access the internal state variables defined in the struct type.
+
+The `evts` parameter is a sdk-managed batch of events to be used for creating new events. For that, the SDK uses a slab allocator and reuses the same event batch at every iteration to improve performance. The lenght of the `evts` list represents the maximum size of each event batch.
+Each element of the batch is an instance of `sdk.EventWriter` that provides handy methods to write the event info and data. Event data can be written with the Go `io.Writer` interface.
```
mstemm, line 508: I think it might be useful to document and describe the EventWriter(s) interfaces (and maybe EventReader too) as a section "Interacting with events". You could talk about the slab allocator and how it's very memory efficient, as long as someone uses the interfaces to access events.

content/en/docs/plugins/developers_guide.md
```diff
@@ -0,0 +1,1081 @@
...
+			return 0, err
+		}
+	}
+	return n, nil
+}
+```
+
+#### Printing Events As Strings
+
+Source plugins must define a `String()` method to format the contents of events created with a previous call to `NextBatch()`. The event data is available through an `io.ReadSeeker` interface.
```
mstemm, line 550: I'd reiterate that we use the io.ReadSeeker interface to avoid data copies and keep the common paths fast.

mstemm (commented):
Looks strange to add a review to my own PR, but this is for the plugin-sdk-go and dummy go plugin walkthrough sections, which were written by @jasondellaluce / @leogr .

ykcilborw:
I would be interested in more details on when to use a source plugin vs an extractor plugin since the APIs are very similar.

Since the extract interface is available (optionally) to be implemented by the source plugin, it's unclear to me when I should implement an extractor plugin and when I should just implement the extract method for a source plugin

content/en/docs/plugins/developers_guide.md
leogr: I can't entirely agree with that. I believe we shouldn't hide the fact the plugin author can implement several interfaces (it's not a single interface), and some of those are optional. Perhaps I would instead remove the comment from there and introduce a different wording for optional interfaces. 

IMHO, saying something like is perfectly acceptable and understandable for a Go developer: 

> An extractor plugin can optionally implement the `Destroyer` interface. If so, `Destroy()` will be called to release resources used by a plugin before it gets destroyed by the framework.

jasondellaluce: I think there's a good value proposition in having some interfaces as optional, so I endorse the comment of @leogr. I think adding a subsection such as "Optional Interface" might be the way to go here.

content/en/docs/plugins/developers_guide.md
jasondellaluce: Yeah, that's a typo. Good catch!

content/en/docs/plugins/developers_guide.md
jasondellaluce: I agree with you, this creates some ambiguity.

leogr:
> Since the extract interface is available (optionally) to be implemented by the source plugin, it's unclear to me when I should implement an extractor plugin and when I should just implement the extract method for a source plugin

A simple answer would be: you should implement the extractor interface whenever you want to export a field (ie. `foo.bar`) that then can be used in Falco conditions and outputs. If you don't want/need that, you can only implement the source interface. In such a case, the user will need another extractor plugin to access the data your plugin is providing. For example, if your source plugin uses JSON as the format for raw event data, a user may use the [json plugin](https://github.com/falcosecurity/plugins/tree/master/plugins/json) to access the event data.

Not sure this is the best answer, just hoping it is understandable :smile_cat:

content/en/docs/plugins/developers_guide.md
mstemm: If you want to keep them optional, that's fine. I see the section on the optional interfaces. But can we also include them in the code blocks?

content/en/docs/plugins/developers_guide.md
mstemm: I see the changes and I think they help. I'm wondering if we should talk about a "a source plugin with the optional extraction features enabled." though. I can't see a good use case where someone would want to inject custom events but would *not* define fields to extract from those events. The only reason I can think of would be where the event payload is json and all field extraction would be handled by the json plugin. But that's not a great rule writing experience, to force the rule author to keep specifying json pointer paths into the payloads.

I would think the more common case for new source plugins is that they would both inject events and define fields. So why not steer the docs in that direction?

content/en/docs/plugins/developers_guide.md
mstemm: thanks, looks better!

content/en/docs/plugins/developers_guide.md
mstemm: The new section looks great, but I would move it into the sdk walkthrough instead of the dummy plugin walkthrough.

content/en/docs/plugins/developers_guide.md
mstemm: Looks great thx!

content/en/docs/plugins/developers_guide.md
jasondellaluce: That's a good point. Will re-elaborate the paragraphs to apply this change.

content/en/docs/plugins/developers_guide.md
mstemm: looks great thx!

content/en/docs/plugins/developers_guide.md
mstemm: thanks! the changes help a lot.

content/en/docs/plugins/developers_guide.md
mstemm: thanks for moving it!

leogr (approved):
/approve

PR #515, opened 2021-11-17
doc(rules): document enabled-only rule definitions

jasondellaluce:
Signed-off-by: Jason Dellaluce <jasondellaluce@gmail.com>

**What type of PR is this?**

/kind content

**Any specific area of the project related to this PR?**

/area documentation

**What this PR does / why we need it**:

This updates the rule docs to follow up the changes introduced in https://github.com/falcosecurity/falco/pull/1775.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

PR #534, opened 2022-01-25
[rule_enable_opt] remove reference to bug in docs

pabloopez:
Signed-off-by: pablopez <pablo.lopezzaldivar@sysdig.com>

**What type of PR is this?**

/kind content

/area documentation

During the last community call. @jasondellaluce mentioned that the bug impacting the `enable` field of falco rules will be addressed in the next release. So, this note in the docs won't be required any more.

None that I am aware of, but it might be of interest in https://github.com/falcosecurity/falco/issues/1857


pabloopez:
duplicated, this as already been addressed in #515

PR #542, opened 2022-02-02
docs(plugins): update plugin documentation, api reference, and developer guide

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind content

/area documentation

This is a first pass to update the contents of the plugins documentation to sync it with the newest changes.

cc @leogr, @mstemm


content/en/docs/plugins/_index.md
```diff
@@ -61,11 +61,11 @@ Source plugin authors *must* register the plugin with the Falcosecurity organiza
 
 ## Plugin Event Sources and Interoperability
 
-Events returned by source plugins have an "event source" which describes the information in the event. This is distinct from the plugin name to allow for multiple kinds of plugins to generate the same kind of events. For example, there might be plugins gke-audit-bridge, eks-audit-bridge, ibmcloud-audit-bridge, etc. that all fetch [K8s Audit](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/) information. The plugins would have different names and IDs but would have the same event source "k8s_audit".
+Events returned by source plugins have an "event source" which describes the information in the event. This is distinct from the plugin name to allow for multiple kinds of plugins to generate the same kind of events. For example, there might be plugins gke-audit-bridge, eks-audit-bridge, ibmcloud-audit-bridge, etc. that all fetch [K8s Audit](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/) information. The plugins would have different names and IDs but would have the same event source k8s_audit.
```
leogr, line 64:
```suggestion
Events returned by source plugins have an "event source" that describes the event's information. This is distinct from the plugin name to allow for multiple plugin implementations to generate the same kind of events. For example, there might be plugins gke-audit-bridge, eks-audit-bridge, ibmcloud-audit-bridge, etc. that all fetch [K8s Audit](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/) information. The plugins would have different names and IDs but would have the same event source k8s_audit.
```

content/en/docs/plugins/_index.md
```diff
@@ -129,17 +131,16 @@ Falco configures plugins via the new "plugins" property in falco.yaml. Here's an
...
+    library_path: libcloudtrail.so
+    init_config: ""
+    open_params: ""
+  - name: json
+    library_path: libjson.so
+    init_config: ""
 
 # Optional
-load_plugins: [aws_cloudtrail]
+load_plugins: [cloudtrail]
```
leogr, line 143:
```suggestion
load_plugins: [cloudtrail, json]
```
Does `cloudtrail` work without `json` ?

content/en/docs/plugins/developers-guide.md
```diff
@@ -189,13 +183,25 @@ type Closer interface {
...
 
+Source and extrator plugins can also optionally implement the `sdk.InitSchema` interface. In that case, `InitSchema()` will be used to to return a schema describing the data expected to be passed as a configuration during the plugin initialization. This follows the semantics documented for [`get_init_schema`](../plugin-api-reference/#const-char-plugin-get-init-schema-ss-plugin-schema-type-schema-type-required-no). Currently, the schema must follow the [JSON Schema specific](https://json-schema.org/), which in Go can also be easily auto-generated with external packages (e.g. [alecthomas/jsonschema](https:/github.com/alecthomas/jsonschema)).
+
+Source plugins can optionally implement the `sdk.OpenParams` interface. In that case, `OpenParams()` will be called before opening the event stream to obtains some suggested values that would valid parameters for `Open()`. For more details, see the documentation of [`list_open_params`](../plugin-api-reference/#const-char-plugin-list-open-params-ss-plugin-t-s-ss-plugin-rc-rc-required-no).
```
leogr, line 203:
```suggestion
Source plugins can optionally implement the `sdk.OpenParams` interface. If requested by the application, the framework may call `OpenParams()` before opening the event stream to obtains some suggested values that would valid parameters for `Open()`. For more details, see the documentation of [`list_open_params`](../plugin-api-reference/#const-char-plugin-list-open-params-ss-plugin-t-s-ss-plugin-rc-rc-required-no).
```

content/en/docs/plugins/developers-guide.md
```diff
@@ -189,13 +183,25 @@ type Closer interface {
...
+}
+
 type Progresser interface {
 	Progress(pState sdk.PluginState) (float64, string)
 }
 ```
 
 Source and extrator plugins can optionally implement the `sdk.Destroyer` interface. In that case, `Destroy()` will be called when the plugin gets destroyed and can be used to release any allocated resource.
 
+Source and extrator plugins can also optionally implement the `sdk.InitSchema` interface. In that case, `InitSchema()` will be used to to return a schema describing the data expected to be passed as a configuration during the plugin initialization. This follows the semantics documented for [`get_init_schema`](../plugin-api-reference/#const-char-plugin-get-init-schema-ss-plugin-schema-type-schema-type-required-no). Currently, the schema must follow the [JSON Schema specific](https://json-schema.org/), which in Go can also be easily auto-generated with external packages (e.g. [alecthomas/jsonschema](https:/github.com/alecthomas/jsonschema)).
```
leogr, line 201:
```suggestion
Source and extrator plugins can also optionally implement the `sdk.InitSchema` interface. In that case, `InitSchema()` will be used to to return a schema describing the data expected to be passed as a configuration during the plugin initialization. This follows the semantics documented for [`get_init_schema`](../plugin-api-reference/#const-char-plugin-get-init-schema-ss-plugin-schema-type-schema-type-required-no). Currently, the schema must follow the [JSON Schema specific](https://json-schema.org/), which in Go can also be easily auto-generated with external packages (e.g. [alecthomas/jsonschema](https:/github.com/alecthomas/jsonschema)).
```

Does the framework call `InitSchema()`  always?

content/en/docs/plugins/developers-guide.md
leogr: Does the framework call `OpenParams()`  always?

content/en/docs/plugins/developers-guide.md
```diff
@@ -459,45 +474,63 @@ func (m *MyPlugin) Info() *plugins.Info {
...
+
 ...
 ```
 
 #### Initializing/Destroying the Plugin
 
 The mandatory `Init()` method serves as an initialization entrypoint for plugins. This is where the user-defined configuration string is passed in by the framework. The internal state of the plugin should be initialized at this level. An error can be returned to abort the plugin initialization.
 
-Defining the `Destroy` method is optional, but can be useful if some resource needs to be released before the plugin gets destroyed.
+Defining the `Destroy()` method is optional, but can be useful if some resource needs to be released before the plugin gets destroyed. The `InitSchema()` methos is optional too, but it allows the framework to automatically parse the init config, thus avoiding the need of doing it manually inside `Init()`.
```
leogr, line 485:
```suggestion
Defining the `Destroy()` method is optional but can be useful if some resource needs to be released before the plugin gets destroyed. The `InitSchema()` method is optional too, but it allows the framework to parse the init config automatically, thus avoiding the need of doing it manually inside `Init()`.
```

leogr: typo and (optional) suggestions

content/en/docs/plugins/plugin-api-reference.md
```diff
@@ -87,12 +87,12 @@ The only time where duplicate sources make sense are when a group of plugins can
 
 ##### `const char* plugin_get_fields() [Required: no]`
 
-This function should return the set of fields supported by the plugin. Remember, a field is a name (e.g. `proc.name`) that can extract a value (e.g. `nginx`) from an event (e.g. a syscall event involving a process). The format is a json string which contains an array of objects. Each object describes one field. Here's an example:
+This function should return the set of fields supported by the plugin. Remember, a field is a name (e.g. `proc.name`) that can extract a value (e.g. `nginx`) from an event (e.g. a syscall event involving a process). The format is a json string, with memory owned by the plugin, which contains an array of objects. Each object describes one field. Here's an example:
```
leogr, line 90:
```suggestion
This function should return the set of fields supported by the plugin. Remember, a field is a name (e.g. `proc.name`) that can extract a value (e.g. `nginx`) from an event (e.g. a syscall event involving a process). The return value is a string whose memory is owned by the plugin. The string is json formatted and contains an array of objects. Each object describes one field. Here's an example:
```

leogr: Not sure my suggestion is good. Just a personal preference. Feel free to ignore my suggestion :)

content/en/docs/plugins/plugin-api-reference.md
```diff
@@ -130,7 +158,7 @@ When managing plugin-level state, keep the following in mind:
...
 * On failure, the plugin framework will *not* call plugin_destroy, so do not return any allocated state.
 
-The format of the config string is entirely determined by the plugin author, and is passed unchanged from Falco/the application using the plugin framework to the plugin. Given that, semi-structured formats like JSON/YAML are preferable to free-form text.
+The format of the config string is entirely determined by the plugin author, and by default is passed unchanged from Falco/the application using the plugin framework to the plugin. However, semi-structured formats like JSON/YAML are preferable to free-form text. In those cases, the plugin author can provide a schema describing the config string contents by implementing the optional `get_init_schema` function. If so, the `init()` function can assume the passed-in configuration string to always be well-formed, and can avoid performing any error handling. The plugin framework will take care of automatically parsing it against the provided schema and generating ad-hoc errors accordingly. Please refer to the documentation of `get_init_schema` for more details.
```
leogr, line 161: Shall we recommend JSON?

content/en/docs/plugins/plugin-api-reference.md
```diff
@@ -184,7 +212,7 @@ The struct has the following members:
 * `datalen`: pointer to a 32bit integer. The plugin will set it the size of the buffer pointed by data.
 * `ts`: the event timestamp, in nanoseconds since the epoch. Can be (uint64_t)-1, in which case the engine will automatically fill the event time with the current time.
 
-It is not necessary to fill in the evtnum struct member when returning events via plugin_next/plugin_next_batch. Event numbers are assigned by the plugin framework.
+It is not necessary to fill in the evtnum struct member when returning events via plugin_next_batch. Event numbers are assigned by the plugin framework.
```
leogr, line 215: Is that just unnecessary? Or the plugin author must NOT do that?

leogr (commented):
Great improvements and a lot of fixes, thank you!

I have just left some suggestions (mostly nitpicking :smile_cat: ), but this PR seems already good.
:+1:

content/en/docs/plugins/_index.md
mstemm: Well the plugin operates but the rules in aws_cloudtrail_rules.yaml won't. I think we should include both to be a realistic example.

content/en/docs/plugins/_index.md
```diff
@@ -186,12 +187,7 @@ If you are interested in authoring your own plugin, or modifying an existing plu
 
 ## Plugins
 
-Plugins authored by the Falcosecurity community are at the [plugins github repository](https://github.com/falcosecurity/plugins). The current plugins are:
-
-* **cloudtrail**: Reads Cloudtrail JSON logs from files/S3 and injects as events.
-* **dummy**: Reference plugin use to document plugins interface.
-* **dummy_c**: Like dummy, but written in C++.
-* **json**: Extracts values from any JSON payload.
```
mstemm, line 194: This is a bit minor, but I actually like listing the plugins here, as it gives a reader a flavor of what's available without having to follow the link. I don't feel super strongly about it, though.

mstemm (dismissed):
This looks good to me. One minor suggestion, and all of @leogr 's suggested changes all look worthwhile as well.

I can re-approve quickly when ready. Thanks!

content/en/docs/plugins/_index.md
jasondellaluce: I think we might want to auto-generate this from the plugin registry in the future. Plus, we may also want to re-invent the https://falco.org/docs/event-sources/ section a bit, now that we have plugins. This is more a reminder for ourselves, so for now I'm gonna restore this list!

content/en/docs/plugins/developers-guide.md
jasondellaluce: If implemented, yes. The way the Go SDK does this is by always implementing a default `plugin_get_init_schema` that returns no schema (and thus is ignored by the framework), which eventually calls `InitSchema()` if the plugin struct implements it.

content/en/docs/plugins/developers-guide.md
jasondellaluce: Same as `InitSchema()`, see the comment above. However, this suggestion is correct because, unlike `get_init_schema`, `list_open_params` is only executed by the framework if requested by the `libs` application.

content/en/docs/plugins/plugin-api-reference.md
jasondellaluce: I think it is pretty much recommended right now. By using JSON, you have access to the schema feature using JSON Schema, which delegates the configuration validation to the framework. I think this the `get_init_schema` doc section make this explicit, but if you want I can add few lines here too.

content/en/docs/plugins/plugin-api-reference.md
jasondellaluce: More that unnecessary, I would say useless. The value would be overwritten even if the plugin writes into it.

leogr (approved):
/approve

:partying_face:

PR #563, opened 2022-03-07
fix(install): add the helm command to use leastPrivileged and PSP

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

The GKE helm command did not include an option to configure Pod Security Policy. This feature has already been deprecated by Kubernetes and will be removed in version 1.25, but for now it is still needed in GKE, otherwise you will see messages like:

```
PodSecurityPolicy: unable to admit pod: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[1]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[2]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[3]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[4]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[5]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[6]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[7]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
```

Fixes #

### DIFF
--- a/content/en/docs/getting-started/third-party/production.md
+++ b/content/en/docs/getting-started/third-party/production.md
@@ -55,8 +55,8 @@ to the file:
 Environment='FALCO_BPF_PROBE=""'
 ```
 
-If you are [installing Falco with Helm](https://falco.org/docs/getting-started/third-party/install-tools/#helm), you will need to set the `ebpf.enabled` option to `true`:
+If you are [installing Falco with Helm](https://falco.org/docs/getting-started/third-party/install-tools/#helm), you will need to set the `ebpf.enabled` option and `leastPrivileged` mode to `true`, and for Kubernetes versions earlier than 1.25 enabling the Falco pod security policy is also needed:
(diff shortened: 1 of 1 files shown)


LucaGuerra:
My bad, that is not the default configuration but may only happen with old clusters with deprecated configurations.

PR #570, opened 2022-03-17
docs(content): add a new step in the upgrading phase

Andreagit97:
Signed-off-by: Andrea Terzolo <andrea.terzolo@polito.it>

**What type of PR is this?**

/kind content

/area documentation

When installing Falco with  `.deb`, `.rpm` or with the binary package, we have to remove the old kernel module injected in the kernel, before upgrading Falco. Otherwise, we will have errors like the following, due to driver versions incompatibility:

```
Runtime error: Kernel module does not support PPM_IOCTL_GET_API_VERSION. Exiting.
```
This happens with Falco `0.31.1` because we introduced a mechanism to detect incompatible driver versions. The simple command added here should solve issues with local package installations quoted here :point_down: 

https://github.com/falcosecurity/falco/issues/1941#issuecomment-1068976133

### DIFF
--- a/content/en/docs/getting-started/upgrade.md
+++ b/content/en/docs/getting-started/upgrade.md
@@ -8,6 +8,12 @@ This section provides upgrading paths for Falco if previously installed followin
 
 ## Upgrading
 
+According to the installation method you chose, you first have to remove the active kernel module before upgrading Falco to the latest version:
+
+```shell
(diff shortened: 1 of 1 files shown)


leogr (approved):
/approve

jasondellaluce (approved):
/approve

PR #578, opened 2022-04-22
docs: add stdout buffering explanation

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

Add more explicit documentation on the behavior of the `-U/--unbuffered` option.
I think we need more documentation on that because in some clusters events might take a long time to appear in `kubectl logs -f` because of log output buffering. Although that is not the way people use to get falco logs in production it is the way some of us use to check if Falco is working, to do debugging, or for training scenarios. If you don't know about buffering that looks like a Falco bug (while it isn't) and sometimes issues are opened for that ( e.g. https://github.com/falcosecurity/falco/issues/1957 )

Fixes #

### DIFF
--- a/content/en/docs/alerts/_index.md
+++ b/content/en/docs/alerts/_index.md
@@ -30,6 +30,12 @@ Standard output is useful when using Fluentd or Logstash to capture logs from co
 
 When run in the background via the `-d/--daemon` command line option, standard output messages are discarded.
 
+### Standard Output buffering
+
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):
/approve 

Thanks Luca!

PR #579, opened 2022-04-26
fix: updating eBPF probe minimum kernel version guidance

abroglesc:
/kind cleanup

/area blog

Updates minimum kernel version guidance to be 4.14. As per https://github.com/falcosecurity/libs/blob/master/driver/bpf/quirks.h#L14-L16 4.14 is the minimum kernel version


Andreagit97 (approved):
Thank you @abroglesc! LGTM!

PR #586, opened 2022-05-11
Added notes about hugo version and gRPC with SELinux

josevnz:
**What type of PR is this?**

/kind content

**Any specific area of the project related to this PR?**

> /area documentation

**What this PR does / why we need it**:

* Added note about interaction of Falco gRPC publish and SELinux
* Added note on site compilation to mention that Hugo 0.90 will not work

**Which issue(s) this PR fixes**:

Fixes #585 

**Special notes for your reviewer**:

PR #588, opened 2022-05-14
SELinux gRPC startup error notes

josevnz:
Signed-off-by: Jose Vicente Nunez <josevnz@kodegeek.com>

**What type of PR is this?**

/kind content

/area documentation

Added note about an error that prevents gRPC server from being started if SELinux is enabled. Mitigation for such situation is also described.

Fixes #586 

New pull request, previous one had too many issues that prevented proper merging into master.

### DIFF
--- a/content/en/docs/alerts/_index.md
+++ b/content/en/docs/alerts/_index.md
@@ -178,3 +178,41 @@ Here's the same output, pretty-printed:
 ## gRPC Output
 
 If you'd like to send alerts to an external program connected via gRPC API (for example, the [falco-exporter](https://github.com/falcosecurity/falco-exporter)), you need to enable both the `grpc` and `grpc_output` options as described under the [gRPC Configuration section](/docs/grpc/#configuration).
+
+
+{{% pageinfo color="primary" %}}
+
+_SELinux note_: If you have [SELinux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux) enabled you may see the following error while starting Falco with gRPC support enabled:
+
+```shell
+
+May 08 12:14:41 myserver falco[15442]: Falco version X.XX.X (driver version ZZZZ)
+May 08 12:14:41 myserver falco[15442]: Falco initialized with configuration file /etc/falco/falco.yaml
+May 08 12:14:41 myserver falco[15442]: Loading rules from file /etc/falco/falco_rules.yaml:
+May 08 12:14:42 myserver falco[15442]: Loading rules from file /etc/falco/falco_rules.local.yaml:
+May 08 12:14:42 myserver falco[15442]: Loading rules from file /etc/falco/k8s_audit_rules.yaml:
+May 08 12:14:43 myserver falco[15442]: Starting internal webserver, listening on port 8765
+May 08 12:14:43 myserver falco[15442]: gRPC server threadiness equals to 2
+May 08 12:14:43 myserver falco[15442]: grpc: {"created":"@1652026483.513407096","description":"No address added out of total 1 resolved","file":"/build/release/grpc-prefix/src/grpc/src/core/e>
+May 08 12:14:43 myserver falco[15442]: Error starting gRPC server
+```
+
+You may want to run SELinux in 'permissive' mode while you figure out how to give Falco the proper permission to create the gRPC server on startup:
+
+```shell
+sudo setenforce 0
+```
+
+To make this permanent across reboots (if you have Grubby installed):
+
+```shell
+sudo grubby --default-kernel
+/boot/vmlinuz-5.6.13-100.fc30.x86_64
+sudo grubby --update-kernel=/boot/vmlinuz-5.6.13-100.fc30.x86_64 --args='selinux=1 enforcing=0'
+sudo reboot -n
+```
+
+Then check your system logs (for example using ```journalctl --no-page --since 'YYYY-MM-DD' --grep selinux```) and check the messages.
+
+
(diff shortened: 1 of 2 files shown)


josevnz:
Hello @jasondellaluce ,

It is OK to merge this pull request? Let me know if you still see issues.

Thanks!

content/en/docs/alerts/_index.md
```diff
@@ -178,3 +178,41 @@ Here's the same output, pretty-printed:
...
+```
+
+To make this permanent across reboots (if you have Grubby installed):
+
+```shell
+sudo grubby --default-kernel
+/boot/vmlinuz-5.6.13-100.fc30.x86_64
+sudo grubby --update-kernel=/boot/vmlinuz-5.6.13-100.fc30.x86_64 --args='selinux=1 enforcing=0'
+sudo reboot -n
+```
```
leogr, line 213: I'd give more generic instruction or pointers. IMO, the Grubby example may be confusing for non-Grubby users.

leogr (commented):
Thank you!

I just left a comment, otherwise SGTM :)

content/en/docs/alerts/_index.md
josevnz: Hi. Hmmm, let me add some code with a in-place sed.

josevnz: Added the changes, let me know what you think.

content/en/docs/alerts/_index.md
```diff
@@ -178,3 +178,62 @@ Here's the same output, pretty-printed:
...
+May 08 12:14:43 myserver falco[15442]: Error starting gRPC server
+```
+
+You may want to run SELinux in 'permissive' mode while you figure out how to give Falco the proper permission to create the gRPC server on startup:
+
+```shell
+sudo setenforce 0
+```
+
+To make this permanent across reboots (if you have Grubby installed):
```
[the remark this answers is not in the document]
josevnz: Hello,

My steps are to let SELinux run on "Permissive" mode, which will allow the application run but it will log what SELinux considers is an inappropriate behavior; Troubleshooting SELinux is a time consuming issue and sadly I don't have enough cycles right now on how to write a step to step tutorial how to add the rules to make it work (in reality I haven't seen the first large company that uses SELinux properly, that's why Falco is so  attractive).

Maybe for another update on the documentation? I can change the language explaining than using the permissive mode will allow the user to explore the issue and works toward a solution if SELinux is a must.

If not, let me know and I will retract this pull request.

Thanks.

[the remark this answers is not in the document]
LucaGuerra: Thanks Jose. First of all I agree with you that the change to the documentation is valuable, so no need to retract the PR :rocket: ! The error message is confusing and knowing that it may be a SELinux issue is a great thing to have in the manual, thanks a lot for documenting it. It also makes sense that adding something more detailed can be done in another PR if you, or any other SELinux expert, has time to write it.

Here's what I have in mind: instead of adding the code snippet to disable SELinux permanently, have the pointers to the official documentation so users are well informed about what they're doing. I'll send you a suggestion snippet to show you what I mean more precisely.

content/en/docs/alerts/_index.md
```diff
@@ -178,3 +178,62 @@ Here's the same output, pretty-printed:
...
+#     targeted - Targeted processes are protected,
+#     minimum - Modification of targeted policy. Only selected processes are protected. 
+#     mls - Multi Level Security protection.
+SELINUXTYPE=targeted
+
+sudo reboot -n
+```
+
+Then check your system logs (for example using ```journalctl --no-page --since 'YYYY-MM-DD' --grep selinux```) and check the messages.
+
```
LucaGuerra, line 237:
```suggestion

Then restart Falco to verify that it works as expected. To troubleshoot the issue you can then review your system logs (for example using ```journalctl --no-page --since 'YYYY-MM-DD' --grep selinux```). Please refer to the [SELinux manual](https://github.com/SELinuxProject/selinux-notebook/blob/main/src/auditing.md#auditing-selinux-events) for more information about how to read the audit log and correct your SELinux configuration. When done, run:

```shell
sudo setenforce 1
```
to restore enforcing mode.

It is also possible to keep the permissive configuration across reboots if this is necessary to troubleshoot the issue. More information can be found by looking at the examples in the [configuration section in the SELinux manual](https://github.com/SELinuxProject/selinux-notebook/blob/main/src/global_config_files.md#etcselinuxconfig).
```

LucaGuerra: ^ here's the suggestion I'm referring to. Note that `sudo setenforce 1` should be in a "shell" snippet exactly like `sudo setenforce 0` but GitHub doesn't let me do that that from a suggestion snippet :man_shrugging:

josevnz:
I realized that something else is going on this system; I disabled SELinux (after trying your suggestions to reproduce the problem) and the issue shows up after the process is restarted like this:

```shell
[josevnz@macmini2 ~]$ getenforce 
Disabled
[josevnz@macmini2 ~]$ sudo -i
sudo -i
[root@macmini2 ~]# systemctl restart falco.service 
[root@macmini2 ~]# 
Broadcast message from systemd-journald@macmini2 (Sat 2022-06-18 14:56:23 EDT):

falco[1433]: Error starting gRPC server
````

So my documentation update is completely pointless as is trying to solve a problem that doesn't exist with SELinux.

I will open a issue if I think this is a bug, but most likely a configuration issue.

PR #595, opened 2022-05-18
Add exceptions.md Chinese file

cytianying:
Add exceptions.md Chinese file

Signed-off-by: tianying <tianying@uniontech.com>


content/zh/blog/exceptions.md
```diff
@@ -0,0 +1,181 @@
+---
+标题: Falco 规则现在支持例外
```
Issif, line 2: Don't translate the metadata keys please

cytianying: OK, I'll revise it.

PR #701, opened 2022-10-13
update(sources/drivers): refresh `Falco Drivers` page with updated info

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind content

/area documentation

This PR tries to refresh the content of the `Falco Drivers` page with up-to-date info 

### DIFF
--- a/content/en/docs/event-sources/drivers.md
+++ b/content/en/docs/event-sources/drivers.md
@@ -1,5 +1,5 @@
 ---
-title: Falco Drivers
+title: Syscall Events
 weight: 1
 ---
 
--- a/content/en/docs/event-sources/drivers.md
+++ b/content/en/docs/event-sources/drivers.md
@@ -3,15 +3,18 @@ title: Falco Drivers
 weight: 1
 ---
 
-Falco depends on a driver that taps into the stream of system calls on a machine and passes those system calls to user space.
+Falco uses different instrumentations to analyze the system workload and pass security events to userspace. We usually refer to these instrumentations as **syscall sources** since the generated events are strictly related to the syscall context.
 
-The kernel module called `falco` is the default driver. Alternatively, an eBPF probe can be used.
+There are three supported syscall sources:
 
-The Falco project has three different kind of drivers.
-
-- Kernel module
+- Kernel module *(dafault)*
 - eBPF probe
 - Userspace instrumentation
+
+|             | Kernel module | eBPF probe | Userspace       |
+| ----------- | ------------- | ---------- | --------------- |
+| **x86_64**  | >= 2.6        | >= 4.14    | No requirements |
+| **aarch64** | >= 3.4        | >= 4.17    | No requirements |
 
 ## Kernel module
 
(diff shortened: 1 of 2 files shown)


content/en/docs/event-sources/drivers.md
```diff
@@ -3,55 +3,60 @@ title: Falco Drivers
...
-Falco depends on a driver that taps into the stream of system calls on a machine and passes those system calls to user space.
+Falco uses different instrumentations to analyze the system workload and pass security events to userspace. We usually refer to these instrumentations as **syscall sources** since the generated events are strictly related to the syscall context.
 
-The kernel module called `falco` is the default driver. Alternatively, an eBPF probe can be used.
+There are three supported syscall sources:
 
-The Falco project has three different kind of drivers.
-
-- Kernel module
+- Kernel module *(dafault)*
```
vjjmiras, line 10: default

vjjmiras, line 10:
```suggestion
- Kernel module *(default)*
```

vjjmiras:
Despite the little correction I did, lgtm.

Andreagit97:
ei @vjjmiras thank you very much for this! I would propose also changing the page title from `Falco Drivers` to `Syscall Events` since the section is called `Event Sources`, WDYT? @jasondellaluce

content/en/docs/event-sources/drivers.md
Andreagit97: ty

vjjmiras:
What is actually the source these events are retrieved from, the syscalls or the kernel?
I was thinking of **Kernel Events**, but that might not be correct, so **Syscalls (Linux Kernel)** could be a better title.

Andreagit97:
Good question, the source of these events is the kernel for `ebpf` and `kernel module` but this is not true for the userspace instrumentation, so probably we cannot use `Kernel Events`. I was thinking about `Syscall events` because all these events are generated when we enable Falco with the so-called `syscall source` and moreover we already have a dedicated section called [Actions for dropped System call events](https://falco.org/docs/event-sources/dropped-events/)

Probably `System call Events` is even better to be compliant with the aforementioned section...

content/en/docs/event-sources/drivers.md
```diff
@@ -1,57 +1,62 @@
...
 To install the eBPF probe, please refer to the [installation](/docs/getting-started/installation/#install-driver) page.
 
-To enable the eBPF support in Falco set the `FALCO_BPF_PROBE` environment variable to an empty value (ie. `FALCO_BPF_PROBE=""`) or otherwise explicitly set it to the path where the eBPF probe resides.
+To enable the eBPF support in Falco set the `FALCO_BPF_PROBE` environment variable to an empty value (ie. `FALCO_BPF_PROBE=""`, in this way Falco will search under `$HOME/.falco` directory for a file called `falco-bpf.o`) or otherwise explicitly set it to the path where the eBPF probe resides.
 
 ## Userspace instrumentation
 
-Differently from the other drivers, as the name suggests, userspace instrumentations happens 100% in userspace.
+{{% pageinfo color="warning" %}}
+This [syscall source](https://github.com/falcosecurity/pdig) is archived. Falco still has the support for it but the userspace instrumentation is no more maintained.
```
jasondellaluce, line 36:
```suggestion
[The pdig project](https://github.com/falcosecurity/pdig) is currently in archived status. Falco still has support for userspace instrumentation, but pdig is not actively maintained currently.
```

content/en/docs/event-sources/drivers.md
```diff
@@ -1,57 +1,62 @@
 ---
-title: Falco Drivers
+title: Syscall Events
```
jasondellaluce, line 2: For all's worth, I think a good name could be "**Kernel Events**" here. We don't just push syscalls, but also things like container metadata, events from various tracepoints etc. Plus, the term "Kernel" also gathers in its umbrella concepts such as gVisor, pdig, or trace files, which don't really install a driver but still deal with events coming from the kernel. What do you think? @Andreagit97 @vjjmiras

vjjmiras: What I hear is that we need to distinguish between **Events** and **Event sources**. Currently we use use different sources (kernel module, eBPF, pdig, gVisor) to obtain the same kind of events (syscalls). 
Hence, **Kernel Events** would be partly incorrect, for not being a source, but partly correct, for grouping these different sources (since having separate sections might not be worth it).

Andreagit97: That's a good point, you bought me I will change it @vjjmiras @jasondellaluce

jasondellaluce: In the falco terminology, a **source** is a stream of events with a certain structure. For example, the `syscall` source is composed of events collected from the kernel, and can be provided by the kmod, eBPF, gVisor, and userspace instrumentation. Another example is the `k8s_audit` source, which is composed of Kubernetes Audit Events. You could say that an event source is characterized by the events that get streamed through it.

As such, kmod/eBPF/gVisor are not sources, but just a way to instrument the kernel to obtain events.

Andreagit97: In the end `Kernel Events` seems reasonable for this context

leogr (approved):

PR #751, opened 2022-11-29
refactor /docs/rules

vjjmiras:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

Fixes #


jasondellaluce (commented):
This looks amazing! Few things that we need to highlight on rules:
- Visibility is defined in cascade and is quite important:
  - A macro can only reference macros defined before it
  - A list can only reference lists defined before it
  - A macro can reference any list
  - A rule can reference any macro
- Override is complete, there is no "merging": when I re-define a rule, it will entirely replace the previous one, so if I define less YAML map fields than required Falco can return an error. Only exceptions are "enabled", that if defined alone simply enables/disables a previously-defined rule, and "append", which for both macros and rules just appends the condition/exceptions field.
- Not sure if I missed this, but we should mention that rule condition fields can also be of "list type", meaning that if of that type Falco extracts a list of values for that field instead of a single value, which can only be used with list-compatible operators such as `in`, `intersects`, and `contains`

vjjmiras:
@jasondellaluce From your comments, can you review these new sections?
* https://deploy-preview-751--falcosecurity.netlify.app/docs/rules/basic-elements/#visibility
* https://deploy-preview-751--falcosecurity.netlify.app/docs/rules/appending/#rewriting-rules

leogr:
Hey @vjjmiras could you rebase please?

@jasondellaluce PTAL

vjjmiras:
Thanks, @leogr, and @jasondellaluce. I believe it's fixed now.

leogr (approved):

content/en/docs/rules/versioning.md
```diff
@@ -0,0 +1,34 @@
...
+### Falco Rules File Versioning
+
+The Falco rules files included with Falco include a new top-level object, `required_engine_version: N`, that specifies the minimum engine version required to read this rules file. If not included, no version check is performed when reading the rules file. Here's an example:
+
+```yaml
+# This rules file requires a falco with falco engine version 7.
+- required_engine_version: 7
+```
+
+If a rules file has an `engine_version` greater than the Falco engine version, the rules file is loaded and an error is returned.
```
Issif, line 34: is loaded or not loaded?

PR #816, opened 2023-02-14
fix: improve Debian instructions

sgaist:
**What type of PR is this?**

> /kind bug

 /kind cleanup

> /kind design

> /kind user-interface

> /kind content

> /kind translation

> /area blog

/area documentation

> /area videos

This PR improves the Debian installation instructions as they are currently relying on a deprecated and soon to be removed tool (apt-key).

The refactored instructions follow best practice in terms of apt source handling to improve the system security as well as working as a power user rather than root.

Nothing particular

### DIFF
--- a/content/en/docs/getting-started/installation.md
+++ b/content/en/docs/getting-started/installation.md
@@ -74,13 +74,13 @@ Let's see an example of how to install the package in a Debian-like system, for 
 3. Update the package list
 
     ```bash
-    sudo apt-get update -y
+    ```
     
     > In older versions of Debian (those previous to Debian 10), you might need to additionally install the package `apt-transport-https` to allow access to the Falco repository.
     > To do so, you can run the following command:
     >
     > sudo apt-get install apt-transport-https
-    ```
+    
 
 4. Install some required dependencies that are needed to build the kernel module and the BPF probe
 
--- a/content/en/docs/getting-started/installation.md
+++ b/content/en/docs/getting-started/installation.md
@@ -74,6 +74,7 @@ Let's see an example of how to install the package in a Debian-like system, for 
 3. Update the package list
 
     ```bash
+    sudo apt-get update -y
     ```
     
     > In older versions of Debian (those previous to Debian 10), you might need to additionally install the package `apt-transport-https` to allow access to the Falco repository.
--- a/content/en/docs/getting-started/installation.md
+++ b/content/en/docs/getting-started/installation.md
@@ -75,6 +75,11 @@ Let's see an example of how to install the package in a Debian-like system, for 
 
     ```bash
     sudo apt-get update -y
+    
+    > In older versions of Debian (those previous to Debian 10), you might need to additionally install the package `apt-transport-https` to allow access to the Falco repository.
+    > To do so, you can run the following command:
+    >
+    > sudo apt-get install apt-transport-https
     ```
 
 4. Install some required dependencies that are needed to build the kernel module and the BPF probe
--- a/content/en/docs/getting-started/installation.md
+++ b/content/en/docs/getting-started/installation.md
@@ -62,29 +62,29 @@ Let's see an example of how to install the package in a Debian-like system, for 
 1. Trust the `falcosecurity` GPG key
 
     ```bash
-    curl -s https://falco.org/repo/falcosecurity-packages.asc | apt-key add -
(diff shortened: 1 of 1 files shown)


Issif:
Can you rebase your PR please, and sorry for the delay, we're in a full revamp of the website.

sgaist:
Sure thing and no worries

vjjmiras:
@sgaist The new procedure to add a key, do you know from which Debian version it is compatible?

sgaist:
@vjjmiras AFAIK, the support for `signed-by` is available since apt 1.1.

The key management instructions here are valid for at least Debian 7 (at the time of writing).
With Debian 6, you start getting in trouble because of TLS but you can still copy the file over and continue from there. There will be other warnings related to public key availability though.
That said, Falco cannot be installed as is on these old version of versions of Debian because of the base requirements it has.

One side question, should there be instructions to install `apt-transport-https` ? Beside curl and gpg, this package is mandatory to access the Debian server.

vjjmiras (changes requested):
@sgaist, I've been testing the changes in other versions of Debian and you're right, the command works, but there are bigger issues to be concerned about, like the expired keys for the repositories.

Regarding the `apt-transport-https`, I found it is not necessary for Debian 10 and newer distributions, so I added a little change to incorporate your suggestion.

sgaist:
@vjjmiras about the key expiration, AFAIK, you will get an error message when the key has expired whether you use the new or old way to do handle them.

Do you mean you would add a note telling the user that they have to update the key when expiration happens ?

content/en/docs/getting-started/installation.md
```diff
@@ -62,37 +62,42 @@ Let's see an example of how to install the package in a Debian-like system, for
...
     ```bash
-    echo "deb https://download.falco.org/packages/deb stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list
+    echo "deb [signed-by=/usr/share/keyrings/falco-archive-keyring.gpg] https://download.falco.org/packages/deb stable main" | sudo tee -a /etc/apt/sources.list.d/falcosecurity.list
     ```
 
 3. Update the package list
 
     ```bash
-    apt-get update -y
+    sudo apt-get update -y
```
vjjmiras, line 77:
```suggestion
    ```
```

content/en/docs/getting-started/installation.md
```diff
@@ -62,37 +62,42 @@ Let's see an example of how to install the package in a Debian-like system, for
...
+    
+    > In older versions of Debian (those previous to Debian 10), you might need to additionally install the package `apt-transport-https` to allow access to the Falco repository.
+    > To do so, you can run the following command:
+    >
+    > sudo apt-get install apt-transport-https
     ```
```
vjjmiras, line 83:
```suggestion
```

content/en/docs/getting-started/installation.md
```diff
@@ -62,37 +62,42 @@ Let's see an example of how to install the package in a Debian-like system, for
...
 3. Update the package list
 
     ```bash
-    apt-get update -y
+    sudo apt-get update -y
+    
+    > In older versions of Debian (those previous to Debian 10), you might need to additionally install the package `apt-transport-https` to allow access to the Falco repository.
+    > To do so, you can run the following command:
+    >
+    > sudo apt-get install apt-transport-https
```
vjjmiras, line 82:
```suggestion
    > sudo apt-get install apt-transport-https
    
```

vjjmiras (changes requested):
I just realized I left some quotes open, my bad. The suggestions should fix it.

vjjmiras:
> @vjjmiras about the key expiration, AFAIK, you will get an error message when the key has expired whether you use the new or old way to do handle them.
> 
> Do you mean you would add a note telling the user that they have to update the key when expiration happens ?

No, I just meant that if they still maintain so older systems, they have bigger concerns than how the repository key is deployed :-)

sgaist:
> I just realized I left some quotes open, my bad. The suggestions should fix it.

Almost ! :-D

> > > @vjjmiras about the key expiration, AFAIK, you will get an error message when the key has expired whether you use the new or old way to do handle them.
> > >
> > Do you mean you would add a note telling the user that they have to update the key when expiration happens ?
> 
> No, I just meant that if they still maintain so older systems, they have bigger concerns than how the repository key is deployed :-)

Ok, now I see your point :-)

vjjmiras (approved):
LGTM

PR #963, opened 2023-05-12
fix #791 - json_output option influences --version output

### LINKED ISSUE #791: (docs) json_output option influences `--version` output
ref: https://falco.org/docs/reference/daemon/config-options/

After https://github.com/falcosecurity/falco/pull/2351 has been merged in, `json_output=true` now influences the output of `falco --version`. As I commented in https://github.com/falcosecurity/falco/pull/2351#issuecomment-1397402538, this was surprising to me since it didn't use affect `--version`.

> This was surprising to me initially when pulling in the latest upstream changes because I expected json_output to only influence the output of the falco_events / alerts / logs; I did not expect this option to influence the output of --version.
>
> (My confusion was also compounded a user runs falco --version and doesn't have permissions to read /etc/falco/falco.yaml, so their output appeared as what I was familiar with whereas a script executing w/ the right context/permissions had the output in JSON.)

/area documentation

**What would you like to be added**:

* additional note that json_output influences the output of `falco --version`

**Why is this needed**:

* can become confusing when scripts parse the output (driver, engine versions, etc.) of `--version` and were not expecting JSON (which was the old behavior)

Happy-Dude:
Add a clarifying blurb for `json_output` config option

**What type of PR is this?**

> /kind content

> /area documentation

Fixes #791

### DIFF
--- a/data/en/reference/daemon/config_options.yaml
+++ b/data/en/reference/daemon/config_options.yaml
@@ -56,7 +56,7 @@
 - name: json_output
   type: Boolean
   description: |
-    If `true` (default is `false`), print falco alert messages and rules file loading/validation results as json, which allows for easier consumption by downstream programs. Default is `false`.
+    If `true` (default is `false`), print falco alert messages and rules file loading/validation results as json, which allows for easier consumption by downstream programs. Also changes falco `--version` CLI argument output as json.
 
 - name: json_include_output_property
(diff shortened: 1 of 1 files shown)


vjjmiras (approved):
LGTM

Issif (approved):

Issue #972, opened 2023-05-23
Migrate to Google Analytics 4 (GA4)


chalin:
This issue is being opened in the context of the following request: https://cncfservicedesk.atlassian.net/browse/CNCFSD-1770

### Existing website info

- UA site tag: [`UA-40398182-19`][ua]
  - Insert link to **Google UA console here**.
  - TBD if there are other UA site tags being used (e.g., a separate tag for blog entries)
- Netlify hosted (does CNCF have access?)

---

This issue is part of a CNCF-wide effort to upgrade project websites to GA4 since **Google has deprecated Universal Analytics** (UA). For more details, see:

- https://github.com/cncf/techdocs/issues/108, including reference to a suggested [migration process](https://github.com/cncf/techdocs/blob/main/docs/analytics.md#migrating-to-google-analytics-4), which we've used to [migrate a few CNCF projects](https://docs.google.com/spreadsheets/d/1Mx4LhdI2Un-rvGMI73SlHxQH9D2HABAJclMB3dd6lnA) already.

### Tasks

There are quite a few ways that we migrate to GA4. The following is, in most cases, the simplest:

- [x] Create a GA4 site tag under the **CNCF projects** account.
  The new GA4 stream measurement ID is: **[`G-D5DEHT7DJ1`][ga4]**
- [x] [Connect][] the UA ID from the GA4 site tag
- [x] Configure GA4 website ID ([gtag.js]) through [Netlify snippet injection], "Google Analytics 4 - in production only", just before `</head>`:
  ```liquid
  {% if CONTEXT == 'production' %}
  <script async src="https://www.googletagmanager.com/gtag/js?id=G-D5DEHT7DJ1"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
  
    gtag('config', 'G-D5DEHT7DJ1');
  </script>
  {% endif %}
  ```
- [x] Confirm that the GA4 site tag is receiving events
- [x] #973
- [x] Link GA4 ID to search console

If it is still relevant for the UA site tag to receive events, then:

- [x] Link the UA and GA4 site tags (in either direction, which ever is most convenient)
- [x] Confirm that the UA site tag is still receiving events

### Google GA4 property created in 2022

[...]

PR #973, opened 2023-05-23
Drop UA ID

chalin:
Contributes to:

- #972

/area documentation
/kind cleanup

/cc @LucaGuerra @leogr

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -127,8 +127,9 @@ anchor = "smart"
 
 [services]
 [services.googleAnalytics]
-# Comment out the next line to disable GA tracking. Also disables the feature described in [params.ui.feedback].
-id = "UA-40398182-19"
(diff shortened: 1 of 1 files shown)


chalin:
Marking as draft until we resolve the duplicate GA4-property issue. For context, see https://cncfservicedesk.atlassian.net/browse/CNCFSD-1770?focusedCommentId=24180.

chalin:
Via Slack, @aijamalnk wrote:

> ... we will be migrating to a new GA4 owned by CNCF.

Thanks for the confirmation. In that case, we can proceed with the review and merge of this PR. Removing draft status.

chalin:
This PR went live about 25 min ago (at the time of writing). I confirm that the **UA ID is still receiving events**, as it should, since it is connected from the new CNCF GA4 ID:

> image

PR #1010, opened 2023-06-20
Document fs.path.* fields

mstemm:
Add a new page that describes how fs.path.* fields work, including the specific syscalls that are supported, which fields are mapped to fs.path.name/source/target, and an example rules file that takes advantage of the fields.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind translation

> /area blog

/area documentation

> /area videos

Fixes #

Corresponding code change PR: https://github.com/falcosecurity/libs/pull/1060

### DIFF
--- a/content/en/docs/rules/fspath.md
+++ b/content/en/docs/rules/fspath.md
@@ -48,47 +48,39 @@ The below tables show:
 
 #### Single Argument File System Syscalls
 
-| Syscall  | Falco Libs Event Identifier   | `fs.path.name` field | From Enter Event?|
-| -------  | ------------------------------| ---------------------|------------------|
-| mkdir    | `PPME_SYSCALL_MKDIR`          | `evt.rawarg.path`    | Yes              |
-| mkdir    | `PPME_SYSCALL_MKDIR_2`        | `evt.rawarg.path`    | No               |
-| mkdirat  | `PPME_SYSCALL_MKDIRAT`        | `evt.rawarg.path`    | No               |
-| rmdir    | `PPME_SYSCALL_RMDIR`          | `evt.rawarg.path`    | Yes              |
-| rmdir    | `PPME_SYSCALL_RMDIR_2`        | `evt.rawarg.path`    | No               |
-| unlink   | `PPME_SYSCALL_UNLINK`         | `evt.rawarg.path`    | Yes              |
-| unlink   | `PPME_SYSCALL_UNLINK_2`       | `evt.rawarg.path`    | No               |
-| unlinkat | `PPME_SYSCALL_UNLINKAT`       | `evt.rawarg.name`    | Yes              |
-| unlinkat | `PPME_SYSCALL_UNLINKAT_2`     | `evt.rawarg.name`    | No               |
-| open     | `PPME_SYSCALL_OPEN`           | `evt.rawarg.name`    | No               |
-| openat   | `PPME_SYSCALL_OPENAT`         | `evt.rawarg.name`    | Yes              |
-| openat   | `PPME_SYSCALL_OPENAT_2`       | `evt.rawarg.name`    | No               |
-| openat2  | `PPME_SYSCALL_OPENAT2`        | `evt.rawarg.name`    | No               |
-| fchmod   | `PPME_SYSCALL_FCHMOD`         | `fd.name`            | No               |
-| fchmodat | `PPME_SYSCALL_FCHMODAT`       | `evt.rawarg.filename`| No               |
-| chmod    | `PPME_SYSCALL_CHMOD`          | `evt.rawarg.filename`| No               |
-| chown    | `PPME_SYSCALL_CHOWN`          | `evt.rawarg.path`    | No               |
-| lchown   | `PPME_SYSCALL_LCHOWN`         | `evt.rawarg.path`    | No               |
-| fchown   | `PPME_SYSCALL_FCHOWN`         | `fd.name`            | No               |
-| fchownat | `PPME_SYSCALL_FCHOWNAT`       | `evt.rawarg.pathname`| No               |
-| quotactl | `PPME_SYSCALL_QUOTACTL`       | `evt.rawarg.special` | No               |
-| umount   | `PPME_SYSCALL_UMOUNT`         | `evt.rawarg.name`    | No               |
-| umount   | `PPME_SYSCALL_UMOUNT_1`       | `evt.rawarg.name`    | No               |
-| umount2  | `PPME_SYSCALL_UMOUNT2`        | `evt.rawarg.name`    | No               |
+| Syscall  | `fs.path.name` field |
+| -------  | ---------------------|
+| mkdir    | `evt.rawarg.path`    |
+| mkdirat  | `evt.rawarg.path`    |
+| rmdir    | `evt.rawarg.path`    |
+| unlink   | `evt.rawarg.path`    |
+| unlinkat | `evt.rawarg.name`    |
+| open     | `evt.rawarg.name`    |
+| openat   | `evt.rawarg.name`    |
+| openat2  | `evt.rawarg.name`    |
+| fchmod   | `fd.name`            |
+| fchmodat | `evt.rawarg.filename`|
+| chmod    | `evt.rawarg.filename`|
+| chown    | `evt.rawarg.path`    |
+| lchown   | `evt.rawarg.path`    |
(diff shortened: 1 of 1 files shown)


content/en/docs/rules/fspath.md
```diff
@@ -0,0 +1,107 @@
...
+| mkdirat  | `PPME_SYSCALL_MKDIRAT`        | `evt.rawarg.path`    | No               |
+| rmdir    | `PPME_SYSCALL_RMDIR`          | `evt.rawarg.path`    | Yes              |
+| rmdir    | `PPME_SYSCALL_RMDIR_2`        | `evt.rawarg.path`    | No               |
+| unlink   | `PPME_SYSCALL_UNLINK`         | `evt.rawarg.path`    | Yes              |
+| unlink   | `PPME_SYSCALL_UNLINK_2`       | `evt.rawarg.path`    | No               |
+| unlinkat | `PPME_SYSCALL_UNLINKAT`       | `evt.rawarg.name`    | Yes              |
+| unlinkat | `PPME_SYSCALL_UNLINKAT_2`     | `evt.rawarg.name`    | No               |
+| open     | `PPME_SYSCALL_OPEN`           | `evt.rawarg.name`    | No               |
+| openat   | `PPME_SYSCALL_OPENAT`         | `evt.rawarg.name`    | Yes              |
+| openat   | `PPME_SYSCALL_OPENAT_2`       | `evt.rawarg.name`    | No               |
```
incertum, line 58: Explain to the reader why we sometimes have 2? When I write a rule with openat syscall which one is used? Same applies to the other instances.

content/en/docs/rules/fspath.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Motivation
+
+A variety of different syscalls take file system paths as arguments. However, there is little consistency in the fields that can access those file system paths. Depending on the syscall, the file system path might be available in `evt.rawarg.path`, `evt.rawarg.pathname`, `evt.rawarg.name`, `fd.name`, etc. Depending on the specific libs event identifier (this generally changes over time based on the falco version, or based on the libc version), the field might differ. And for some system calls, the file system path is actually in the enter event instead of the exit event. This makes it very difficult to write simple Falco rules that act on file system paths, because the field must depend on the syscall as well as the event direction.
+
+To help address this inconsistency, in Falco version XXX we added a new set of fields that normalize file system paths across a variety of syscalls.
+
+### `fs.path.*` fields
+
+The following fields are available for any syscall that operates on a file system path:
```
incertum, line 20: We had a discussion about this before, perhaps state that this does not apply to when spawning a process that uses an executable on disk, aka proc.exepath is kept solely in the proc field class for xyz reasons ...

mstemm:
Do we know which falco version will contain the corresponding libs change? If so, I can put the falco version in the docs now (currently XXX).

mstemm:
Replaced xxx placeholder with 0.36.

incertum:
> Do we know which falco version will contain the corresponding libs change? If so, I can put the falco version in the docs now (currently XXX).

Good point, perhaps since we are just now introducing the feature start with the today and then it should just be the most recent PPME event and not an older one. Going forward adding version information is an excellent idea!

Also @mstemm one more thought we have this general Falco fields reference doc, and this one is a separate doc, could we still add `fs.*` field class to the other one and link to this new page, WDYT?

jasondellaluce:
/hold

Up until Falco 0.36

content/en/docs/rules/fspath.md
mstemm: It was kind of covered in the introduction, but I moved it to the table introduction just as background.

mstemm:
> > Do we know which falco version will contain the corresponding libs change? If so, I can put the falco version in the docs now (currently XXX).
> 
> Good point, perhaps since we are just now introducing the feature start with the today and then it should just be the most recent PPME event and not an older one. Going forward adding version information is an excellent idea!
> 
> Also @mstemm one more thought we have this general Falco fields reference doc, and this one is a separate doc, could we still add `fs.*` field class to the other one and link to this new page, WDYT?

The page https://falco.org/docs/reference/rules/supported-fields/ is actually semi-automatically generated from `falco --list --markdown`. I'm not sure if that's fully automated or not. I kind of thought it was handled as a step of creating a new falco release. I'm not sure on that, though.

incertum:
> > > Do we know which falco version will contain the corresponding libs change? If so, I can put the falco version in the docs now (currently XXX).
> > 
> > 
> > Good point, perhaps since we are just now introducing the feature start with the today and then it should just be the most recent PPME event and not an older one. Going forward adding version information is an excellent idea!
> > Also @mstemm one more thought we have this general Falco fields reference doc, and this one is a separate doc, could we still add `fs.*` field class to the other one and link to this new page, WDYT?
> 
> The page https://falco.org/docs/reference/rules/supported-fields/ is actually semi-automatically generated from `falco --list --markdown`. I'm not sure if that's fully automated or not. I kind of thought it was handled as a step of creating a new falco release. I'm not sure on that, though.

I don't know either, while it is automated maybe the paragraphs on top are still edited here right? Anyhow that's probably where we would want to cross-link to this dedicated page.

content/en/docs/rules/fspath.md
```diff
@@ -0,0 +1,113 @@
...
+
+All of these convert relative paths to absolute paths when needed, using the thread's current working directory (cwd).
+
+`fs.path.nameraw/fs.path.sourceraw/fs.path.targetraw` are like the above but do *not* convert relative paths to absolute paths. They always contain the original path, which may or may not be relative.
+
+The fields only work for exit events and only return a value if the syscall was successful.
+
+The below tables show:
+* the specific syscalls that are are supported
+* the specific falco event identifers are supported. The reason there are multiple event identifers for the same syscall (e.g. MKDIR vs MKDIR_2) is that libs used to define new events every time we added/modified arguments to the enter or exit event. Older applications using the older version of libs will use the older event identifier for the syscall name, while newer applications will use the newer event identifier.
```
incertum, line 45: This is great it explains the difference between enter and exit events, and I am still not sure if we have to list the older event types in the table since we start introducing this with Falco 0.36. Going forward it makes sense to version this table. Alternatively it would even be ok to remove the column "Falco Libs Event Identifier" altogether as perhaps it confuses most Falco end users more than it helps. Falco adopters primarily think of the syscall string name as event I think and libs adopters likely are familiar with the source code and consult the event table. No real opinion here, just thinking out loud.

mstemm: We can remove it if it's not valuable. I think it's valuable to devs but not users.

incertum: Maybe it's hard to keep that page updated with changing PPME codes? Let's remove it, we can rather add a link to the event_table source code?

content/en/docs/rules/fspath.md
```diff
@@ -0,0 +1,113 @@
...
+* fs.path.name
+* fs.path.nameraw
+* fs.path.source
+* fs.path.sourceraw
+* fs.path.target
+* fs.path.targetraw
+
+`fs.path.name` is for file operations that work on a path like open, unlink, rmdir, etc. For other file operations that have a source and target, like cp, symlink, link, mv, etc, there are fields `fs.path.source` and `fs.path.target`.
+
+All of these convert relative paths to absolute paths when needed, using the thread's current working directory (cwd).
```
nate-double-u, line 37:
```suggestion
These convert relative paths to absolute paths when needed, using the thread's current working directory (cwd).
```

content/en/docs/rules/fspath.md
```diff
@@ -0,0 +1,113 @@
...
+
+`fs.path.nameraw/fs.path.sourceraw/fs.path.targetraw` are like the above but do *not* convert relative paths to absolute paths. They always contain the original path, which may or may not be relative.
+
+The fields only work for exit events and only return a value if the syscall was successful.
```
nate-double-u, line 41:
```suggestion
The fields only work for exit events and only return a value if the syscall succeeds.
```

nate-double-u (commented):
A copy-edit pass. Feel free to take or leave any of the suggestions.

content/en/docs/rules/fspath.md
mstemm: Ok I removed the PPME codes as well as the column that denoted whether the value comes from the enter or exit event.

PR #1015, opened 2023-06-22
new(reference): add some documents regarding Falco capabilities

Andreagit97:
**What type of PR is this?**

/kind content

/area documentation

This document explains which capabilities are necessary to run Falco with various drivers


content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Capabilities
+description: Capabilities required to run Falco
+weight: 20
+---
+
+To analyze kernel events we have [3 possible sources](/docs/event-sources/kernel/): kernel module, eBPF probe, modern eBPF probe.
```
vjjmiras, line 7:
```suggestion
To analyze kernel events we have [3 possible means](/docs/event-sources/kernel/): the kernel module, the eBPF probe, and the modern eBPF probe.
```

vjjmiras: Should we start calling the eBPF probe, "the classic eBPF probe"?

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Capabilities
```
vjjmiras, line 2:
```suggestion
title: Linux Capabilities
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+description: Capabilities required to run Falco
```
vjjmiras, line 3:
```suggestion
description: Linux Capabilities required to run Falco
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+weight: 20
+---
+
+To analyze kernel events we have [3 possible sources](/docs/event-sources/kernel/): kernel module, eBPF probe, modern eBPF probe.
+
+The kernel module can run only in privileged mode while the 2 eBPF probes could also use capabilities to run with fewer privileges.
```
vjjmiras, line 9:
```suggestion
The kernel module can run only in privileged mode while the eBPF probes could leverage the Linux capabilities to run with fewer privileges.
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+## eBPF probe
+
+The minimal set of capabilities required by Falco to run the eBPF probe is the following:
```
vjjmiras, line 13:
```suggestion
The minimal set of capabilities required by Falco to run the eBPF probe are the following:
```

vjjmiras: Again, calling it "the eBPF probe" leaves me in state where I don't know which eBPF probe we are talking about.

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+- `CAP_SYS_ADMIN`
+- `CAP_SYS_RESOURCE`
+- `CAP_SYS_PTRACE`
+
+### Explanation
+
+- **CAP_SYS_RESOURCE**: Falco needs this to call the `setrlimit` syscall. This syscall is used together with `RLIMIT_MEMLOCK` flag to change the amount of memory that can be mlocked into RAM. The default memory limit value is very low, so even a very simple eBPF program will fail. The workaround is to increase the default value to something acceptable so that eBPF maps can be correctly mlocked in memory.
```
vjjmiras, line 21:
```suggestion
- **CAP_SYS_RESOURCE**: Falco needs this capability to be able to call the `setrlimit` syscall. The `setrlimit` syscall is used together with the `RLIMIT_MEMLOCK` flag to change the amount of memory that can be _mlocked_ into RAM. The default value for this memory limit is very low, so even a very simple eBPF program would fail. The workaround is to increase the default value to something acceptable so eBPF maps can be correctly _mlocked_ in memory.
```

vjjmiras: so even a very simple eBPF program ... would fail or could fail?

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+- **CAP_SYS_PTRACE**: Falco needs this capability because it accesses fields like environ in the proc file system. From the userspace standpoint, the permission to do so is mapped to the `CAP_SYS_PTRACE` capability. For the curious reader, see [environ_open](https://elixir.bootlin.com/linux/latest/source/fs/proc/base.c#L937) implementation in the kernel.
```
vjjmiras, line 22: fields like environ or paths like /proc/environ?

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+- **CAP_SYS_ADMIN**: Falco needs this capability to load eBPF programs, maps and interact with the system using the `bpf` syscall.
+
+This set of capabilities should always work but under some conditions is possible to replace the `CAP_SYS_ADMIN` with 2 less intrusive capabilities: `CAP_SYS_BPF` and `CAP_SYS_PERFMON`.
```
vjjmiras, line 25:
```suggestion
This set of capabilities should work most of the time but under some conditions, it is possible to replace the `CAP_SYS_ADMIN` with two more granular capabilities: `CAP_SYS_BPF` and `CAP_SYS_PERFMON`.
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+The conditions to satisfy are the following:
+
+1. A kernel version that supports these capabilities. `5.8` is the first kernel version that officially supports them but as always they could be also backported on older versions.
```
vjjmiras, line 29:
```suggestion
1. A kernel version that supports these capabilities. The Linux Kernel version `5.8` is the first one that officially supports them but they could have been backported on older versions on some distributions.
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+2. An acceptable value of `kernel.perf_event_paranoid` config. Reading the [manual](https://linuxsecurity.expert/kb/sysctl/kernel_perf_event_paranoid/) it seems that `perf_event_paranoid` influences only the behavior of unprivileged users, but under the hood, some distros like Debian and Ubuntu introduce other `perf_event_paranoid` levels. Consider [`Ubuntu`](https://kernel.ubuntu.com/git/ubuntu-stable/ubuntu-stable-jammy.git/tree/kernel/events/core.c#n11991) as an example:
```
vjjmiras, line 31:
```suggestion
2. An acceptable value of `kernel.perf_event_paranoid` config. Reading the [manual](https://linuxsecurity.expert/kb/sysctl/kernel_perf_event_paranoid/) it seems that `perf_event_paranoid` influences only the behavior of unprivileged users, but under the hood, some distributions like Debian or Ubuntu introduce additional `perf_event_paranoid` levels. Consider [`Ubuntu`](https://kernel.ubuntu.com/git/ubuntu-stable/ubuntu-stable-jammy.git/tree/kernel/events/core.c#n11991) as an example:
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+	if (perf_paranoid_any() && !capable(CAP_SYS_ADMIN))
+	return -EACCES;
+
+	// where perf_paranoid_any is:
+	static inline bool perf_paranoid_any(void) {
+	return sysctl_perf_event_paranoid > 2;
+	}
+	```
+
+ 	As you can easily notice if your `kernel.perf_event_paranoid` is `>2` `CAP_PERFMON` will be not enough, you will need `CAP_SYS_ADMIN`!
```
vjjmiras, line 43:
```suggestion
 	As you can notice, when your `kernel.perf_event_paranoid` is `>2` the capability `CAP_PERFMON` won't suffice, you would still need `CAP_SYS_ADMIN`.
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+ 	So before disabling `CAP_SYS_ADMIN` check your `perf_event_paranoid` value with `sysctl kernel.perf_event_paranoid` and ensure this is compatible with your distro enforcement.
```
vjjmiras, line 44:
```suggestion
 	So before disabling `CAP_SYS_ADMIN` check your `perf_event_paranoid` value with `sysctl kernel.perf_event_paranoid` and make sure their values are compatible with your distribution enforcement.
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Modern eBPF probe
+
+The minimal set of capabilities required by Falco to run the modern eBPF probe is the following:
```
vjjmiras, line 48:
```suggestion
The minimal set of capabilities required by Falco to run the modern eBPF probe are the following:
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+- `CAP_SYS_BPF`
+- `CAP_SYS_PERFMON`
+- `CAP_SYS_RESOURCE`
+- `CAP_SYS_PTRACE`
+
+### Explanation
+
+No particular explanation is needed the capabilities are the same explained in the previous section. Here there are no issues with `kernel.perf_event_paranoid` so this set of capabilities should always work.
```
vjjmiras, line 57:
```suggestion
The mentioned capabilities require no further explanation since they were already discussed in detail in the previous section. This set of capabilities should ways work since here there are no issues with `kernel.perf_event_paranoid`.
```

content/en/docs/reference/capabilities/_index.md
```diff
@@ -0,0 +1,59 @@
...
+
+> **Please note**: we will try to do our best to keep this as the minimum required set but due to [some issues with CO-RE relocations](https://lore.kernel.org/bpf/CAGQdkDvYU_e=_NX+6DRkL_-TeH3p+QtsdZwHkmH0w3Fuzw0C4w@mail.gmail.com/T/#u) is possible that in the future this could change
```
vjjmiras, line 59:
```suggestion
> **Please note**: we will try to do our best to keep this as the minimum required set but due to [some issues with CO-RE relocations](https://lore.kernel.org/bpf/CAGQdkDvYU_e=_NX+6DRkL_-TeH3p+QtsdZwHkmH0w3Fuzw0C4w@mail.gmail.com/T/#u) it is possible that this changes in the future.
```

vjjmiras (commented):
I'm not sure this should be considered a reference, but rather I'd include it [here](https://falco.org/docs/event-sources/kernel/). WDYT?

content/en/docs/reference/capabilities/_index.md
Andreagit97: Here i left "is" since the subset should be "set"

content/en/docs/reference/capabilities/_index.md
Andreagit97: would fail, thank you!

content/en/docs/reference/capabilities/_index.md
Andreagit97: paths like /proc/environ, thanks!

content/en/docs/reference/capabilities/_index.md
Andreagit97: As before i kept "is"

Andreagit97:
I renamed the eBPF probe to "classic eBPF probe" as you suggested, and moved the section in "Kernel sources" let me know what you think :)

vjjmiras:
I think it looks much better there, thanks for considering it.

Some things I'd still change:

In the first table:
  - Uppercase to the probe's names, like in the list
  - Using the /docs/event-sources/drivers/ slug doesn't seem to help when using the internal headers as links. Should we maybe update it to /docs/events/sources/kernel/?
  -  I'd say the recommended version for the modern probe is 5.8, but there are some exceptions, for consistency with the rest of the table. The exceptions can be found under requirements, as you already pointed.

Two more points from a general perspective:

- What about moving the modern eBPF section to appear before the classic eBPF section? That should be the preferred option when the kernel allows it, right? Or is it still a bit too early?
- We've been missing for a long time a section discussing which driver to choose. This article would be a great place to address that topic, but it doesn't need to be right now.

Andreagit97:
> I think it looks much better there, thanks for considering it.
> 
> Some things I'd still change:
> 
> In the first table:
> 
>     * Uppercase to the probe's names, like in the list

Done :) 

> 
>     * Using the /docs/event-sources/drivers/ slug doesn't seem to help when using the internal headers as links. Should we maybe update it to /docs/events/sources/kernel/?

Done :) thanks

>     * I'd say the recommended version for the modern probe is 5.8, but there are some exceptions, for consistency with the rest of the table. The exceptions can be found under requirements, as you already pointed.

yeah, the only downside is that there are really corner cases in which even 5.8 is not enough so probably I would leave "minimal set of features" because it's the most reliable info we can provide, WDYT?

> 
> Two more points from a general perspective:
> 
>     * What about moving the modern eBPF section to appear before the classic eBPF section? That should be the preferred option when the kernel allows it, right? Or is it still a bit too early?

I would wait at least the next release in the end this is the first release in which it's not experimental

>     * We've been missing for a long time a section discussing which driver to choose. This article would be a great place to address that topic, but it doesn't need to be right now.

That's a fair point but i would keep it for another PR, moreover in my opinion we could also remove the `userspace` docs part since it is no more maintained :thinking:

vjjmiras (approved):
LGTM

PR #1023, opened 2023-06-30
Added KubeSphere Logo as Adopter

halil-bugol:
KubeSphere is a distributed operating system for cloud-native application management, using Kubernetes as its kernel. It provides a plug-and-play architecture, allowing third-party applications to be seamlessly integrated into its ecosystem.

Our version 4.0 is coming soon and we will offer Falco integration directly from the interface in this version. We also use Falco in our own dev and prod environments.

/kind design


halil-bugol:
@Issif I closed the previous PR and opened the new one and the problem went away.

Issif (approved):

PR #1068, opened 2023-07-21
new(docs): clarify kernel/userspace api boundary

### LINKED ISSUE #1063: Increase clarity in the API between the kernel space and the user space
**What to document**

The Falco [review from the CNCF TOC](https://docs.google.com/document/d/12l65c6qC91akgFjzw7BM3KdNJAa5LyDG17sdpdfwf3g/edit) highlighted the need to better document how the data flows between user space and kernel space. See Emily's comment:

> Improve the documentation on the communication APIs and define the data being exchanged. It should be written to address both adopters and potential code contributors. Understanding those APIs and the data that flows through them would allow contributors to comprehend the tool and its potential capabilities far better to extend its use as well as assist adopters in understanding what sorts of information can be extracted and reused within their enterprise architectures for business and threat decisions.

Essentially, some parts of the protocol are implementation details, while some other parts can help adopters and contributors in understanding what kind of information is exchanged. I feel like we already have [some of this information](https://falco.org/docs/reference/rules/supported-events/) in our public reference but our adopters are very technical and need to learn more in order to make informed decisions, while allowing a less steep learning curve for contributors willing to extend the system.

LucaGuerra:
**What type of PR is this?**

/kind content
/area documentation

As per title. See issue #1063 for more details.

To summarize:
* The reference is now updated with flags documentation
* Similarly, the webpage explains the difference between types of events
* Kernel event architecture and plugin event architecture has been split
* We documented the user facing version information for kernel API and schema
* Likewise, the internal binary format common to all user<->kernel communication is now documented

Fixes #1063


content/en/docs/event-sources/kernel/architecture.md
```diff
@@ -0,0 +1,68 @@
...
+
+Meaning that its encoding will be composed of an header containing the timestamp and tid, `nparams` will be 4 and the complete encoding will be:
+
+```
+[header] [uint16(8)] [uint16(8)] [uint16(8)] [uint16(32)] [res] [oldfd] [newfd] [flags]
+```
+
+![](/docs/images/kernel_source_capture.png)
+
+All the supported events can be used in Falco rules, can be enabled or disabled if necessary and are documented in the reference.
```
leogr, line 64:
```suggestion
All the supported events can be used in Falco rules, can be enabled or disabled if necessary and are documented in the reference.
```

I don't believe this is true in all cases. There can be internal events that aren't available in the rules.

content/en/docs/event-sources/kernel/architecture.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Kernel Events Architecture
+description: 'Architecture overview for kernel events.'
+linktitle: Kernel Events Architecture Overview
+weight: 20
+---
+
+This document describes the overall architecture that allows events from kernel sources to be ingested by Falco and how to use the libraries to inspect the data collection flow, and how Falco manages the boundary between the kernel and userspace.
```
leogr, line 8: I like this, but we should clarify that these APIs aren't intended for adopters.

These APIs only allow Falco to employ the drivers, so they must be considered for internal use only and are subject to change at any time.

leogr (commented):
Nice explanation!

I just had a quick look and left some minor suggestions.

LucaGuerra:
In addition, I will send a PR to Falco to update the CLI switches to automatically generate the markdown document in the format you see in this PR.

leogr (approved):
Thank you!

PR #1090, opened 2023-07-26
update(docs): update alert forwarding page

LucaGuerra:
/kind cleanup
/area documentation

Update the homepage and forwarding page to make it clearer, plus fix a typo in the alert link.

In addition, this adds an explicit link to the box in the homepage. This way it'll be much easier for users to find the right page that shows integration capabilities.


leogr (approved):
:heart:

PR #1099, opened 2023-07-27
update(supported events, supported fields): link the two pages together, fix ambiguous wording

### LINKED ISSUE #985: Improve documentation around supported falco fields
/area documentation

**What would you like to be added**:

Expand and improve https://falco.org/docs/reference/rules/supported-fields/:

We could add full schemas for each syscall that are accessible via `evt.arg.*`  or at least refer to the source code file https://github.com/falcosecurity/libs/blob/master/driver/event_table.c.

Current mention is not explicit enough, I still don't know what valid field names are as typical end user (unless I am very familiar with the source code and each Linux syscall man page or deduct them from existing Falco rules):
> "Event fields applicable to syscall events. Note that for most events you can access the individual arguments/parameters of each syscall via evt.arg, e.g. evt.arg.filename."

Related to https://github.com/falcosecurity/libs/issues/1134
> Many users are also unfamiliar with the `evt.arg.*` fields and their meaning. Enhancing our documentation can address these issues by providing clear explanations and references to the concept of files in Linux.

On that note `mesos` can be removed as it is deprecated and I am not sure what `span` and `fdlist` classes actually represent and how you would use them in Falco. Can this be documented as well?

**Why is this needed**:

Improve UX for effective use of Falco.

LucaGuerra:
**What type of PR is this?**

/kind cleanup
/area documentation

As written in issue 985, if you land in the supported fields page without noticing its sibling "supported events" page you might be left wondering what the events and the arguments actually are. This PR adds a link to make it clearer. In addition, I noticed that in one page the arguments were called "arguments" or "args" and in the other were called "params". Let's call them "arguments" (or args) in both pages for consistency.

Fixes #985

### DIFF
--- a/content/en/docs/reference/rules/supported-events/index.md
+++ b/content/en/docs/reference/rules/supported-events/index.md
@@ -11,7 +11,7 @@ Here are the system call event types and args supported by the [kernel module an
 Note that several event types exist:
 * [Syscall events](#syscall-events) correspond to Linux system calls. Most of them have parameters, documented below, while some are detected as generic and they only offer the syscall ID.
 * [Tracepoint events](#tracepoint-events) represent internal kernel events that may be significant but don't directly translate to any syscall.
-* [Metaevents](#metaevents) are internal Falco events.
+* [Metaevents](#metaevents) are generated from supplementary data sources, for instance, during data enrichment procedures or when the need for asynchronous actions arises. This group also encompasses some of Falco's internally produced events (such as the `drop` event) that are unavailable for rules.
 * [Plugin events](#plugin-events) are internal Falco events. They act as an envelope for actual plugin event data and are not available to rules. In order to write rules that use plugins use the fields documented in the individual plugin.
 
 <!--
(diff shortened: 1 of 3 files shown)


content/en/docs/reference/rules/supported-events/index.md
```diff
@@ -11,7 +11,7 @@ Here are the system call event types and args supported by the [kernel module an
 Note that several event types exist:
 * [Syscall events](#syscall-events) correspond to Linux system calls. Most of them have parameters, documented below, while some are detected as generic and they only offer the syscall ID.
 * [Tracepoint events](#tracepoint-events) represent internal kernel events that may be significant but don't directly translate to any syscall.
-* [Metaevents](#metaevents) are internal Falco events. They are generally not available for rules.
+* [Metaevents](#metaevents) are internal Falco events.
```
leogr, line 14:
```suggestion
* [Metaevents](#metaevents) are generated from supplementary data sources, for instance, during data enrichment procedures or when the need for asynchronous actions arises. This group also encompasses some of Falco's internally produced events (such as the `drop` event) that are unavailable for rules.
```

IIRC, only `EC_INTERNAL` are filtered out. Any other metaevent goes thru the rule engine as common events.

Issif (approved):

PR #1180, opened 2023-10-12
exlude all non-english pages from the search results

Issif:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind user-interface

> /kind content

/kind translation

> /kind event

/area blog

/area documentation

/area community

This PR removes from the search results all pages which are not in En. it avoids to have duplicated contents in the results.

Fixes #


LucaGuerra:
Is there a way to make the search language aware? So if I have the language set to Japanese and I type something I get only the Japanese pages etc?

Issif:
> Is there a way to make the search language aware? So if I have the language set to Japanese and I type something I get only the Japanese pages etc?

I searched for and nothing like that in [Lunr](https://www.docsy.dev/docs/adding-content/search/#local-search-with-lunr) which is really limited. It also means, cause we don't have all pages translated, the search must fall back on EN pages. Not easy.

leogr (approved):

PR #1264, opened 2024-02-02
update(docs): update build from source page

### LINKED ISSUE #1091: Update info in "Build Falco from source" page
**What is wrong with the website or needs to be cleaned?**

The dependencies list at https://falco.org/docs/getting-started/source/#dependencies page may be not up-to-date.
For example, [civetweb was dropped long time ago](https://github.com/falcosecurity/falco/pull/1952).

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area documentation

Update the build from source documentation to Falco 0.37.0. I have tested many build paths and it looks like they are working. Also, please see inline comments.

Fixes #1091

### DIFF
--- a/content/en/docs/install-operate/source.md
+++ b/content/en/docs/install-operate/source.md
@@ -15,34 +15,31 @@ doing all this, chances that you are willing to contribute are high! Please read
 {{< tabs name="Dependencies" >}}
 {{% tab name="CentOS / RHEL " %}}
 
-CentOS 7 / RHEL 7
+CentOS 8 Stream / RHEL 8
 
 ```bash
-yum install gcc gcc-c++ git make autoconf automake pkg-config patch libtool glibc-static libstdc++-static elfutils-libelf-devel
+dnf install git gcc gcc-c++ git make cmake elfutils-libelf-devel perl-IPC-Cmd
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/install-operate/source.md
```diff
@@ -278,35 +282,8 @@ You can notice this observing that the option `USE_BUNDLED_DEPS` is `OFF` by def
 
 Changing such option to `ON` causes Falco build to bundle all the dependencies statically.
 
-For the sake of completeness this is the complete list of Falco dependencies:
```
LucaGuerra, line 281: This is removed because it's hard to keep up to date and it does not provide much value here; I would prefer we had a better dependency management process upstream instead.

content/en/docs/install-operate/source.md
```diff
@@ -453,36 +428,6 @@ When enabling this you will be able to make the `bpf` target after:
 make bpf
 ```
 
-### Build using falco-builder container
```
LucaGuerra, line 456: This is also removed because it's out of date. If we want to set up a new build container we can document it better, but since the one mentioned here is no longer maintained I would suggest to remove this

content/en/docs/install-operate/source.md
```diff
@@ -508,54 +453,3 @@ sudo ./userspace/falco/falco -c ../falco.yaml -r ../rules/falco_rules.yaml
 ```
 
 By default, falco logs events to standard error.
-
-
-### Run regression tests
```
LucaGuerra, line 513: These are also out of date. We now have a different repo for testing which is used in Falco plus a bunch of unit and integration tests. We can replace this info with an updated version but for now this is just misleading.

leogr (approved):

PR #1301, opened 2024-03-25
update(docs): fixed command for running privileged ebpf falco docker.

terylt:
The existing privileged docker command for running ebpf incorrectly tries to load the kmod, while running falco in ebpf mode.  By adding an environment variable we force the container to load the ebpf probe.  The command was also missing the `falco` cli.

**What type of PR is this?**

 /kind bug

/area documentation

It fixes the docker command for loading the ebpf version of falco.  The current command errors in two ways:
1. Since it doesn't have `falco` in it, docker incorrectly thinks the executable is `-o`
2. The docker container loads the kmod probe, but falco is loaded in `ebpf` mode.

Fixes #


LucaGuerra (approved):
Thank you!

PR #1331, opened 2024-06-18
update: plugin API reference to version 3.6.0

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind user-interface

/kind content

> /kind event

> /area blog

/area documentation

> /area community

Updates the plugin API reference to version 3.6.0.
Updates docs about enabling plugin metrics in Falco.

Fixes #

Please note that this documents features available in Falco 0.38.1, so it should be merged after the Falco release.

### DIFF
--- a/content/en/docs/metrics/falco-metrics.md
+++ b/content/en/docs/metrics/falco-metrics.md
@@ -21,6 +21,7 @@ metrics:
   state_counters_enabled: true
   kernel_event_counters_enabled: true
   libbpf_stats_enabled: true
+  plugins_metrics_enabled: true
   convert_memory_to_mb: true
   include_empty_values: true
 ```
@@ -1258,7 +1259,21 @@ However, following the Prometheus recommendations, there might be some slight di
 
 ## Plugins Metrics
 
-As of Falco 0.38.0, it is not possible to use the metrics feature when running plugins without the syscalls source. Near-term improvements in this regard are tracked in the following [issue](https://github.com/falcosecurity/falco/issues/3194#issuecomment-2111009270).
+Since version 0.38.1, Falco allows plugins to provide their own set of [custom metrics](/docs/reference/plugins/plugin-api-reference/#get-metrics).
+
+In order to enable plugins metrics, the following configuration elements must be set in the [falco.yaml][1]:
+
+```yaml
+metrics:
+  enabled: true
+  ...
+  plugins_metrics_enabled: true
+  ...
+```
+
+Please note that this doesn't provide default metrics about the plugin itself, and it will show metrics only if the plugin is providing them.
+
+Currently, none of the officially maintained plugins provides any metric, but this will change in the future. 
 
 ## Breaking Changes
 
--- a/content/en/docs/metrics/falco-metrics.md
+++ b/content/en/docs/metrics/falco-metrics.md
@@ -1275,6 +1275,8 @@ Please note that this doesn't provide default metrics about the plugin itself, a
 
 Currently, none of the officially maintained plugins provides any metric, but this will change in the future. 
 
+Lastly, please note that as of Falco 0.38.1, it is not possible to fully use the remaining metrics feature when running plugins without the syscalls source. This is due to some regressions that prevent us from serving a small subset of metrics in this scenario. Near-term improvements in this regard are tracked in the following issue. Please also keep in mind that many of the metrics categories are specific to the syscalls source, such as `libbpf_stats_enabled`, `kernel_event_counters_enabled`, or `state_counters_enabled`.
+
 ## Breaking Changes
 
 {{% pageinfo color=info %}}
--- a/content/en/docs/reference/plugins/plugin-api-reference.md
+++ b/content/en/docs/reference/plugins/plugin-api-reference.md
@@ -16,7 +16,7 @@ At a high level, the API functions are grouped as follows:
 * Functions that are commons to all plugins
 * Functions that implement one specific capability
 
-The C header files [plugin_api.h](https://github.com/falcosecurity/libs/blob/0.11.0/userspace/plugin/plugin_api.h) and [plugin_types.h](https://github.com/falcosecurity/libs/blob/0.11.0/userspace/plugin/plugin_types.h) numerate all the API functions and associated structs/types as they are used by the plugins framework. The whole plugin API and the loader used in Falco are implemented in C in a standaline module located inside [falcosecurity/libs/userspace/plugin](https://github.com/falcosecurity/libs/tree/master/userspace/plugin), and can be imported and reused in other projects using the falcosecurity plugin system (e.g. we have a [plugin loader written in Go](https://github.com/falcosecurity/plugin-sdk-go/tree/main/pkg/loader) developed on top of the C one).
+The C header files [plugin_api.h](https://github.com/falcosecurity/libs/blob/0.17.2/userspace/plugin/plugin_types.h) numerate all the API functions and associated structs/types as they are used by the plugins framework. The whole plugin API and the loader used in Falco are implemented in C in a standalone module located inside [falcosecurity/libs/userspace/plugin](https://github.com/falcosecurity/libs/tree/master/userspace/plugin), and can be imported and reused in other projects using the falcosecurity plugin system (e.g. we have a [plugin loader written in Go](https://github.com/falcosecurity/plugin-sdk-go/tree/main/pkg/loader) developed on top of the C one).
 
 Remember, however, that from the perspective of the plugin, each function name has a prefix `plugin_` e.g. `plugin_get_required_api_version`, `plugin_get_name`, etc.
 
@@ -24,7 +24,7 @@ Since [Falco v0.33.0](/blog/falco-0-33-0), some function symbols of **the plugin
 
 ### Plugin API Versioning
 
-**The current version of the plugin API is `3.0.0`**.
+**The current version of the plugin API is `3.6.0`**.
 
 The plugin API is a formal contract between the framework and the plugins, and it is versioned using [semantic versioning](https://semver.org/). The framework exposes the plugin API version it supports, and each plugin expresses a required plugin API version. If the version required by a plugin does not pass the semantic check with the one supported by the framework, then the plugin cannot be loaded. See the section about [`plugin_get_required_api_version`](#get-required-api-version) for more details.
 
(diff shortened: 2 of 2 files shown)


mrgian:
cc @FedeDP @incertum @jasondellaluce

content/en/docs/reference/plugins/plugin-api-reference.md
```diff
@@ -16,15 +16,15 @@ At a high level, the API functions are grouped as follows:
...
+The C header files [plugin_api.h](https://github.com/falcosecurity/libs/blob/0.17.2/userspace/plugin/plugin_types.h) numerate all the API functions and associated structs/types as they are used by the plugins framework. The whole plugin API and the loader used in Falco are implemented in C in a standaline module located inside [falcosecurity/libs/userspace/plugin](https://github.com/falcosecurity/libs/tree/master/userspace/plugin), and can be imported and reused in other projects using the falcosecurity plugin system (e.g. we have a [plugin loader written in Go](https://github.com/falcosecurity/plugin-sdk-go/tree/main/pkg/loader) developed on top of the C one).
```
FedeDP, line 19:
```suggestion
The C header files [plugin_api.h](https://github.com/falcosecurity/libs/blob/0.17.2/userspace/plugin/plugin_types.h) numerate all the API functions and associated structs/types as they are used by the plugins framework. The whole plugin API and the loader used in Falco are implemented in C in a standalone module located inside [falcosecurity/libs/userspace/plugin](https://github.com/falcosecurity/libs/tree/master/userspace/plugin), and can be imported and reused in other projects using the falcosecurity plugin system (e.g. we have a [plugin loader written in Go](https://github.com/falcosecurity/plugin-sdk-go/tree/main/pkg/loader) developed on top of the C one).
```

content/en/docs/metrics/falco-metrics.md
```diff
@@ -1258,7 +1259,21 @@ However, following the Prometheus recommendations, there might be some slight di
...
+metrics:
+  enabled: true
+  ...
+  plugins_metrics_enabled: true
+  ...
+```
+
+Please note that this doesn't provide default metrics about the plugin itself, and it will show metrics only if the plugin is providing them.
+
+Currently, none of the officially maintained plugins provides any metric, but this will change in the future.
```
incertum, line 1276: Add below as paragraph:

Lastly, please note that as of Falco 0.38.1, it is not possible to fully use the remaining metrics feature when running plugins without the syscalls source. This is due to some regressions that prevent us from serving a small subset of metrics in this scenario. Near-term improvements in this regard are tracked in the following [issue]( https://github.com/falcosecurity/falco/issues/3194#issuecomment-2111009270). Please also keep in mind that many of the metrics categories are specific to the syscalls source, such as `libbpf_stats_enabled`, `kernel_event_counters_enabled`, or `state_counters_enabled`.

content/en/docs/reference/plugins/plugin-api-reference.md
```diff
@@ -161,6 +204,46 @@ Currently, the plugin framework only supports the [JSON Schema format](https://j
...
+
+This function returns the pointer to the first element of an array containing plugin-provided custom metrics.
+
+Each element of the array is a `ss_plugin_metric` which is [defined](https://github.com/falcosecurity/libs/blob/0.17.2/userspace/plugin/plugin_types.h#L331-L345) in the plugin API as follows:
+
+```CPP
+typedef struct ss_plugin_metric
+{
+    const char* name; //Opaque string representing the metric name
+	ss_plugin_metric_type type; // Metric type
```
incertum, line 237:
```suggestion
	ss_plugin_metric_type type; // Metric type indicates whether the metric value is monotonic or non-monotonic.
```

content/en/docs/reference/plugins/plugin-api-reference.md
```diff
@@ -161,6 +204,46 @@ Currently, the plugin framework only supports the [JSON Schema format](https://j
...
+	ss_plugin_metric_value value; // Metric numeric value
+	ss_plugin_metric_value_type value_type; // Metric value data type
```
incertum, line 239:
```suggestion
	ss_plugin_metric_value_type value_type; // Metric value data type, e.g. `uint64_t`
```

content/en/docs/reference/plugins/plugin-api-reference.md
```diff
@@ -691,6 +783,52 @@ typedef struct
 } ss_plugin_init_tables_input;
 ```
 
+### Obtaining subtables accessors
+
+Obtaining table accessors is a bit different for subtables. 
+
+Table accessors can be only obtained at initialization time, however tables may be empty at this time, which means subtables may be not existing yet.
```
incertum, line 790:
```suggestion
Table accessors can be only obtained at initialization time, however tables may be empty at this time, which means subtables may be not yet exist.
```

incertum: Could you provide clear examples how to use the new API instead of just providing the source code snippets?

For example how to access the fd table fields and the cmdline args and how is it different from accessing the currently available fields (see below)?

```
Header:
falcosecurity::table_field m_comm;

CPP:
init(falcosecurity::init_input& in)
{
    using st = falcosecurity::state_value_type;
    auto& t = in.tables();
    m_thread_table = t.get_table(THREAD_TABLE_NAME, st::SS_PLUGIN_ST_INT64);
    m_comm = m_thread_table.get_field(t.fields(), "comm", st::SS_PLUGIN_ST_STRING);
...

std::string tstr;
m_comm.read_value(tr, thread_entry, tstr);
```

mrgian: There is already an example that shows how to obtain subtable accessors.
Once you obtained accessors and subtables handles, accessing subtables fields works just like any other regular table.
But I agree on making this more clear by providing another example 👍🏻

leogr (dismissed):
LGTM.

/hold for a 2nd thought from previous reviewers

content/en/docs/reference/plugins/plugin-api-reference.md
incertum: It may seem intuitive to the folks who developed it. For end users who seek to develop a plugin the docs are not sufficient IMO. Instead of source code snippets from libs, what would be better would be clear examples on how to use the API similar to the snippet above.

leogr: I agree with @incertum . 

Just a question: do we want to address this point in this PR? Or do we want to unblock this PR and enhance its content in a follow-up PR? :thinking:

mrgian: The last commit already adds [an example](https://github.com/mrgian/falco-website/blob/plugin-api-docs-3.5.0/content/en/docs/reference/plugins/plugin-api-reference.md#accessing-subtables) that shows how to access subtable fields. 
There is also [this example](https://github.com/mrgian/falco-website/blob/plugin-api-docs-3.5.0/content/en/docs/reference/plugins/plugin-api-reference.md#obtaining-subtables-accessors) that shows how to obtain subtable accessors. 
These two examples are what the user needs to understand how to access subtables IMHO.

Please tell me if I am missing something 🤔

leogr: LGTM. 

@incertum ?

incertum: Hm the links seem to show libs snippet again? And does not appear to be an example of how to use it in a real plugin developed with the plugin SDK in the plugins repo.

Given the below example that shows how I read the `comm` value how do I read `m_args`, and `m_fd_name` values?

```
Header:
falcosecurity::table_field m_comm;

CPP:
init(falcosecurity::init_input& in)
{
    using st = falcosecurity::state_value_type;
    auto& t = in.tables();
    m_thread_table = t.get_table(THREAD_TABLE_NAME, st::SS_PLUGIN_ST_INT64);
    m_comm = m_thread_table.get_field(t.fields(), "comm", st::SS_PLUGIN_ST_STRING);
...

std::string tstr;
m_comm.read_value(tr, thread_entry, tstr);
```

mrgian: @incertum 
🤔 I think you are confusing plugin examples written using the plugin API, with libs snippets.
The purpose of the `Falco Plugins API Reference` page is to document symbols/types of the plugin API which is a plain-C interface.
For this reason, the example I provided shows how to use the plugin API directly without using the C++/Go plugin SDK.

You mentioned accessing the `fd_name` field. The example I provided shows exactly how to do that: [here](https://github.com/mrgian/falco-website/blob/plugin-api-docs-3.5.0/content/en/docs/reference/plugins/plugin-api-reference.md#obtaining-subtables-accessors) it shows how to get the field accessor to that field, and [here](https://github.com/mrgian/falco-website/blob/plugin-api-docs-3.5.0/content/en/docs/reference/plugins/plugin-api-reference.md#accessing-subtables) it shows how to use that accessor to obtain the actual `fd_name` value from one of the subtables.

Again, this example shows how to use the the plugin API and not the plugin SDK.
Documenting the plugin SDK is out of the scope of this website page IMHO. 
Also support for subtables in the C++ plugin SDK has not been merged yet (https://github.com/falcosecurity/plugin-sdk-cpp/pull/38) 

From what I know, the C++ SDK is still an experimental project, it can be subject to breaking changes and it has not been released. For this reason it has no documentation yet, however it will provide [examples](https://github.com/falcosecurity/plugin-sdk-cpp/pull/38/files#diff-c408a815df00653669c8ace735b01a6c91329e5537ab0bf32ca7dbbe8d17d15eR140).

incertum: Hey I am more interested in how to actually use it in a plugin, I added bunch of comments to my open PR https://github.com/falcosecurity/plugins/pull/419#discussion_r1649282173. Appreciate clear examples once the SDK is updated.

mrgian: I'll provide some examples when the C++ SDK is ready :)
But I think it should be addressed in another PR, since the goal of this one is updating the plugin API reference.

leogr: Ok, I'd like to defer this to a follow-up PR and unblock this meantime.

leogr (approved):
/unhold

PR #1353, opened 2024-08-19
fix: update sysdig training links

tembleking:
**What type of PR is this?**

This updates the links for the Sysdig training content. The old links were removed since we switched to a different training platform.

/kind content
/area documentation
/area community


LucaGuerra (approved):
Thank you for fixing this!

PR #1362, opened 2024-09-04
website restructuring [1/N]

leogr:
**What type of PR is this?**
/kind cleanup

/kind design

/kind user-interface

/kind content

/area blog
/area documentation

This is the first step of the action plan described in https://github.com/falcosecurity/falco-website/issues/1365

This PR specifically focuses on reorganizing the first two sections:

- **Getting started** (that was already ok, just minor enhancements)
- **Setup** (previously _Install and Operate_)

In the work of moving content to new places, this PR also anticipates the Developer Guide section which, however, needs more work (this will be addressed in followup PRs).

 

Help with #1365

Some content is still temporary and must be fixed once all the new sections are placed.
Notably, we miss an installation scenario for the non-kernel-instrumentation use case. As a temporary solution, we will forward users to the Plugins section.


content/en/docs/setup/kubernetes.md
```diff
@@ -0,0 +1,68 @@
...
+
+If you wish to upgrade Falco to a new version, you need to find the corresponding version in the [Falco Helm Chart repository](https://github.com/falcosecurity/charts/blob/master/charts/falco) (e.g. `4.8.1` is for Falco `0.38.2` ) then run:
+
+```sh
+helm upgrade falco -n falco --version 4.8.1
+```
+
+{{% pageinfo color="warning" %}}
+To avoid any possible disruption, before upgrading to a new version, consult the [Falco Helm chart Breaking Changes page](https://github.com/falcosecurity/charts/blob/master/charts/falco/BREAKING-CHANGES.md).
+{{% /pageinfo %}}
```
leogr, line 59: fyi @alacuku

content/en/docs/developer-guide/grpc/_index.md
```diff
@@ -2,6 +2,8 @@
 title: gRPC API
 linktitle: gRPC API
 description: Enable and configure the gRPC capabilities of Falco
+aliases:
+- ../../grpc
```
LucaGuerra, line 6: This redirects to https://deploy-preview-1362--falcosecurity.netlify.app/grpc/ instead of https://deploy-preview-1362--falcosecurity.netlify.app/docs/grpc/ (see: https://falco.org/docs/grpc/ )

content/en/docs/setup/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Setup
+description: Learn how to install and configure Falco
+aliases:
+- ../install-operate
```
LucaGuerra, line 5: Ah yes, it's the same issue as gRPC and redirects to top level instead of `docs/install-operate/`

content/en/docs/setup/container.md
```diff
@@ -0,0 +1,291 @@
...
+docker pull falcosecurity/falco-no-driver:latest
+docker run --rm -it \
+           --privileged \
+           -v /var/run/docker.sock:/host/var/run/docker.sock \
+           -v /proc:/host/proc:ro \
+           -v /etc:/host/etc:ro \
+           falcosecurity/falco-no-driver:latest
+```
+
+#### eBPF Probe {#docker-privileged-ebpf}
```
LucaGuerra, line 46: Why modern -> ebpf -> kmod instead of modern -> kmod -> ebpf ? I think we had this discussion, but let's put it in the PR here so we don't forget.

content/en/docs/setup/production-enviroments.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Production Environments
```
LucaGuerra, line 2: I see the point of this page, but I don't like the title. Can we change it to "Specific Environments" or something similar?

content/en/docs/setup/container.md
leogr: I changed my mind a couple of times regarding this. Although I agree we use a different ordering when the automatic selection is enabled, it may be easier for the reader to have both eBPF solutions very close since they have some similarities. Moreover, from an installation point of view, the legacy eBPF sits precisely in the middle between the modern and the kmod.

That said, I don't have strong opinions or compelling motivations. 

I'm ok to switch the order. Before doing that, I just summon @FedeDP since he worked a lot on the automatic selection, and may have opinions.

leogr:
> I believe there are some missing aliases that can be reached via a Google search:
> 
> https://deploy-preview-1362--falcosecurity.netlify.app/docs/install-operate/ https://deploy-preview-1362--falcosecurity.netlify.app/docs/install-operate/third-party/install-tools/
> 
> https://www.google.com/search?q=%22falco.org%2Fdocs%2Finstall-operate%22+-site%3Afalco.org
> 
> Also, there is a gRPC alias issue I think (too many `../`)

Good catch, thank you! I tried to do my best, but it was still manual and error-prone. Fixing.

content/en/docs/setup/container.md
FedeDP: I'd use the same priority-list we used for driver types in Falco, that is: 
* modern_ebpf
* kmod
* ebpf

See https://github.com/falcosecurity/falco/blob/master/scripts/CMakeLists.txt#L51

leogr: ok two vs. one, you won :smile_cat: 

Changing the ordering.

content/en/docs/getting-started/learning-environments.md
Issif, line 1: the table of contents is strange:
image

content/en/docs/setup/production-enviroments.md
Issif: +1 for `Specific Environments` til we get better

content/en/docs/setup/download.md
Issif, line 1: What about moving this page under the Refs section?

Issif (changes requested):

content/en/docs/setup/download.md
leogr: The `Download` page has historically been prominent in the documentation. I wouldn't move it right now. Once the other sections have been moved around, we can still rethink it.

content/en/docs/getting-started/learning-environments.md
leogr: Fixing, Thank you!

leogr: fixed

PR #1467, opened 2025-04-23
ignore and cleanup vim swp files

tuminoid:
In the blog post section, we have one committed swp file already. Clean that off, and add .swp files to gitignore file.

/kind cleanup


leogr (approved):
Good catch, thank you!

PR #1562, opened 2026-06-07
docs: clarify proc.cmdline behavior for pre-execve events

sathvik-shettyy:
Updated the description of the `proc.cmdline` field to clarify its behavior before an `execve()` call completes.

**What type of PR is this?**

/kind content

/area documentation

Adds a clarification to the `proc.cmdline` field documentation explaining that, for events occurring before an `execve()` call completes, the value may differ from the command line later observed through tools such as `ps` or `pstree`.

Related to falcosecurity/falco#3842

**Special notes for reviewer**:

This is a documentation-only change intended to clarify existing behavior. No functional changes were made.


leogr (commented):
Hi @sathvik-shettyy 

Thanks for taking care of this.

However, the source of this message is here: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/sinsp_filtercheck_thread.cpp#L149-L152

and should be fixed there 🙏

sathvik-shettyy:
Thanks for the review and for pointing me to the source.

You're right — the description originates from the libs repository, so the clarification should be made there instead of in the website documentation.

I'll open a corresponding PR against falcosecurity/libs and link it here.

sathvik-shettyy:
Thanks for the review and for pointing me to the source.

I've submitted the clarification in the libs repository, where the proc.cmdline field description originates.

I'll keep the discussion there and can close this website PR if that's preferred.

